PR #6, opened 2015-09-28
Simplify the wait for changes flow

### LINKED ISSUE #5: Broken on mono
After a file change:

image

/cc @victorhurdugaci

davidfowl:
Also fixes #5 

/cc @victorhurdugaci

### DIFF
--- a/src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
+++ b/src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.Dnx.Watcher.Core
 {
--- a/src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
+++ b/src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
@@ -6,7 +6,6 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Framework.Logging;
 
 namespace Microsoft.Dnx.Watcher.Core
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
```diff
@@ -182,41 +181,37 @@ private void AddProjectAndDependeciesToWatcher(IProject project, IFileWatcher fi
...
         }
 
-        private Task<string> WatchForFileChangeAsync(IFileWatcher fileWatcher, CancellationToken cancellationToken)
+        private async Task<string> WatchForFileChangeAsync(IFileWatcher fileWatcher, CancellationToken cancellationToken)
         {
-            return Task.Run(() =>
+            var tcs = new TaskCompletionSource<string>();
+
+            // Review: Do we care about the task being cancelled or just complete?
+            cancellationToken.Register(() => tcs.TrySetResult(null));
```
victorhurdugaci: For the comment: we want to task to complete. If it throws, the exception propagates and the watcher exists

src/Microsoft.Dnx.Watcher.Core/DnxWatcher.cs
```diff
@@ -182,41 +181,37 @@ private void AddProjectAndDependeciesToWatcher(IProject project, IFileWatcher fi
 
             foreach (var projFile in project.ProjectDependencies)
             {
-                //var fullProjectFilePath = Path.Combine(Path.GetDirectoryName(project.ProjectFile), projFile);
```
victorhurdugaci: I think that was a bug in my code, it should actually do the commented line because projects in the lock file are relative to the project's path. If the project being watched references other projects, it might not watch correctly the dependencies of those projects

davidfowl: Figured but wasn't sure. I can fix it if you want.

davidfowl: Or we can just fix the ProjectDependencies to have the full path.

victorhurdugaci:
Looks good. After you address the comments, :shipit:

davidfowl:
Merged

PR #42, opened 2015-12-11
Catch and display project validation exceptions correctly

pakrym:
Issue: https://github.com/aspnet/dnx-watch/issues/20
@victorhurdugaci

### DIFF
--- a/src/Microsoft.Dnx.Watcher.Core/Impl/ProjectProvider.cs
+++ b/src/Microsoft.Dnx.Watcher.Core/Impl/ProjectProvider.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Microsoft.Extensions.CompilationAbstractions;
 
 namespace Microsoft.Dnx.Watcher.Core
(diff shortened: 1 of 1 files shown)


src/Microsoft.Dnx.Watcher.Core/Impl/ProjectProvider.cs
```diff
@@ -12,19 +12,28 @@ public class ProjectProvider : IProjectProvider
...
 
-            errors = null;
-            project = new Project(runtimeProject);
+            try
+            {
+                project = new Project(runtimeProject);
+            }
+            catch (Exception ex)
+            {
+                errors = ex.Message;
```
victorhurdugaci: You should also include messages from the inner exception (if any). Also, you might want to handle `AggregateExceptions` as a special case

pakrym:
:up: :date:

src/Microsoft.Dnx.Watcher.Core/Impl/ProjectProvider.cs
```diff
@@ -43,11 +53,33 @@ private bool TryGetProject(string projectFile, out Runtime.Project project, out
...
             }
 
             project = null;
             return false;
         }
+
+        private string CollectMessages(Exception exception)
+        {
+            var builder = new StringBuilder();
+            builder.Append(exception.Message);
```
victorhurdugaci: Use AppendLine everywhere

victorhurdugaci:
:shipit:

PR #66, opened 2016-04-12
Add a polling watcher

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/55

cc @moozzyk

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/DotnetFileWatcher.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/DotnetFileWatcher.cs
@@ -1,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.Watcher.Core.Internal
+{
+    internal class DotnetFileWatcher : IFileSystemWatcher
+    {
+        private readonly Func<string, FileSystemWatcher> _watcherFactory;
+        private readonly string _watchedDirectory;
+
+        private FileSystemWatcher _fileSystemWatcher;
+
+        private readonly object _createLock = new object();
+
+        public DotnetFileWatcher(string watchedDirectory)
+            : this(watchedDirectory, DefaultWatcherFactory)
+        {
+        }
+
+        internal DotnetFileWatcher(string watchedDirectory, Func<string, FileSystemWatcher> fileSystemWatcherFactory)
+        {
+            if (string.IsNullOrEmpty(watchedDirectory))
+            {
+                throw new ArgumentNullException(nameof(watchedDirectory));
+            }
+
+            _watchedDirectory = watchedDirectory;
+            _watcherFactory = fileSystemWatcherFactory;
+            CreateFileSystemWatcher();
+        }
+
+        public event EventHandler<string> OnFileChange;
+
+        public event EventHandler OnError;
+
+        private static FileSystemWatcher DefaultWatcherFactory(string watchedDirectory)
+        {
+            if (string.IsNullOrEmpty(watchedDirectory))
+            {
+                throw new ArgumentNullException(nameof(watchedDirectory));
+            }
+
+            return new FileSystemWatcher(watchedDirectory);
+        }
+
+        private void WatcherErrorHandler(object sender, ErrorEventArgs e)
+        {
+            // Recreate the watcher
+            CreateFileSystemWatcher();
+
+            if (OnError != null)
+            {
+                OnError(this, null);
+            }
+        }
+
+        private void WatcherRenameHandler(object sender, RenamedEventArgs e)
+        {
+            NotifyChange(e.OldFullPath);
+            NotifyChange(e.FullPath);
+
+            if (Directory.Exists(e.FullPath))
+            {
+                foreach (var newLocation in Directory.EnumerateFileSystemEntries(e.FullPath, "*", SearchOption.AllDirectories))
+                {
+                    // Calculated previous path of this moved item.
+                    var oldLocation = Path.Combine(e.OldFullPath, newLocation.Substring(e.FullPath.Length + 1));
+                    NotifyChange(oldLocation);
+                    NotifyChange(newLocation);
+                }
+            }
+        }
+
+        private void WatcherChangeHandler(object sender, FileSystemEventArgs e)
+        {
+            NotifyChange(e.FullPath);
+        }
+
+        private void NotifyChange(string fullPath)
+        {
+            if (OnFileChange != null)
+            {
+                // Only report file changes
+                OnFileChange(this, fullPath);
+            }
+        }
+
+        private void CreateFileSystemWatcher()
+        {
+            lock (_createLock)
+            {
+                bool enableEvents = false;
+
+                if (_fileSystemWatcher != null)
+                {
+                    enableEvents = _fileSystemWatcher.EnableRaisingEvents;
+
+                    _fileSystemWatcher.EnableRaisingEvents = false;
+
+                    _fileSystemWatcher.Created -= WatcherChangeHandler;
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/DotnetFileWatcher.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.Watcher.Core.Internal
+{
+    internal class DotnetFileWatcher : IFileSystemWatcher
```
muratg: Already in the Internal namespace. Could this be public instead?

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            _pollingThread = new Thread(new ThreadStart(PollingLoop));
+            _pollingThread.IsBackground = true;
+            _pollingThread.Name = nameof(PollingFileWatcher);
+            _pollingThread.Start();
+        }
+
+        public event EventHandler<string> OnFileChange;
+
+#pragma warning disable CS0067 // not used
```
muratg: More info why this is disabled.  I.e. we need it because of the interface, but compiler requires all the events to be used. Or something along those lines :)

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
```diff
@@ -0,0 +1,243 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+
+namespace Microsoft.DotNet.Watcher.Core.Internal
+{
+    internal class PollingFileWatcher : IFileSystemWatcher
+    {
+        // The minimum interval to rerun the scan
+        private static readonly TimeSpan _minRunInternal = TimeSpan.FromSeconds(.5);
```
muratg: Wasn't the plan to make this an environment variable?

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/FileWatcherFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+
+namespace Microsoft.DotNet.Watcher.Core.Internal
+{
+    public static class FileWatcherFactory
+    {
+        public static IFileSystemWatcher CreateWatcher(string watchedDirectory)
+        {
+            var envVar = Environment.GetEnvironmentVariable("USE_POLLING_FILE_WATCHER");
```
glennc: Are we using the same var for the cshtml PR that @pranavkm is doing? If so then `POLLING_FILE_WATCHER` probably isn't accurate. More like `USE_NETWORK_SHARE_FALLBACK` or `USE_NETWORK_SHARE_WATCHER` or something. If they are different keys then this is fine :)

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
glennc: Now that this is inside dotnet-watch shouldn't all the configuration options be passable as both env and command line args?

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/FileWatcherFactory.cs
pranavkm: I was using `ASPNETCORE_POLL_FOR_FILE_CHANGES`. Open to suggestions for alternatives

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/DotnetFileWatcher.cs
victorhurdugaci: No every type in the internal namespace has to be public

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/FileWatcherFactory.cs
victorhurdugaci: It has nothing to do with network share. It's just enables a polling file watcher which can be used for non-network shares. Also, it's not an ASPNETCORE feature (at least in dotnet-watch)

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
victorhurdugaci: The timeout was never said to be configurable by the user.

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
```diff
@@ -0,0 +1,243 @@
...
+            }
+
+            stopwatch.Stop();
+        }
+
+        private void CreateKnownFilesSnapshot()
+        {
+            _knownEntities.Clear();
+
+            ForeachEntityInDirectory(_watchedDirectory, f =>
```
pranavkm: Could we change this design to only look for files we care about? For instance, move the behavior into `https://github.com/aspnet/dotnet-watch/blob/dev/src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs` and add some sort of eventual consistency mechanism to it? Enumerating a project that uses npm could be very expensive with this approach

victorhurdugaci: That's not possible because the file watcher doesn't support globbing patterns. We would need to watch for new files based on them and the globbing framework that we have doesn't support checking a path against a pattern. That's why we have to watch for everything and then check if that path is in the globbing pattern.

victorhurdugaci:
Tested with an app that has 21K watched files. The changes were instant and you couldn't tell without a profiler the difference in time between the polling watcher and the dotnet one. Also, tested on docker and it works with mounted volumes

victorhurdugaci:
@davidfowl watched the 21K files tests and approved it :)

victorhurdugaci:
Updated the test file watcher to use polling

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/FileWatcherFactory.cs
glennc: I know that, I was throwing out suggestions based on the scenario where it makes sense to use this vs anything else. I am not super worried about them being different, lets talk about it in-person tomorrow.

src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
glennc: I don't feel super strongly that it needs to be configurable, but is there a reason you think it shouldn't be possible? The fact that it adds noise to the help for something that nobody will ever change is a fine enough reason to begin with, but is there anything else?

chribben:
Is this feature available in a NuGet package or Docker image somewhere?

victorhurdugaci:
Yes, it is.

Add the `Microsoft.DotNet.Watcher.Tools` tool in project.json:

```
"tools": {
  "Microsoft.DotNet.Watcher.Tools": {
    "version": "1.0.0-*",
    "imports": "portable-net451+win8"
  }
}
```

Then, to enable the polling watcher, set the environment variable `USE_POLLING_FILE_WATCHER=1` before running `dotnet watch`

chribben:
Nice, will try it out!

PR #69, opened 2016-04-12
Fixing tests to use the package that was just built

moozzyk:
Disabling parallel run to prevent random issues
Adding additional details that should help diagnose failures


victorhurdugaci:
:shipit: after you rebase

PR #71, opened 2016-04-13
Fix tests on unix and make tests more reliable on all platforms

victorhurdugaci:
cc @moozzyk


muratg:
Looks like this would increase the test run time a bit. Not sure if there's an alternative though.

:shipit:

moozzyk:
:ship: :it:

PR #76, opened 2016-04-18
Fix the prefix comparison

victorhurdugaci:
There was an incorrect startwith comparison which was causing some projects to not be watched sometimes.

Fixes https://github.com/aspnet/dotnet-watch/issues/75

@moozzyk can you please review?

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/FileWatcher.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/FileWatcher.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/FileWatcher.cs
```diff
@@ -65,7 +68,9 @@ private void AddDirectoryWatcher(string directory)
             newWatcher.OnFileChange += WatcherChangedHandler;
             newWatcher.EnableRaisingEvents = true;
 
-            _watchers.Add(directory, newWatcher);
+            _watchers.Add(
+                EnsureTrailingSlash(directory),
```
victorhurdugaci: This is redundant. WIll remove

moozzyk:
:ship: :it:

PR #107, opened 2016-05-24
Simplify args

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/105

- Make arguments simpler (see readme)
- Remove all watcher arguments
- `dotnet watch` (no args) shows help
- Remove makefile
- Update readme

Please review @moozzyk @glennc


src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -49,47 +46,20 @@ public static int Main(string[] args)
         // To pass "--help" to the app being watched, you must use "--": dotnet watch -- --help
         internal static void SeparateWatchArguments(string[] args, out string[] watchArgs, out string[] appArgs)
```
moozzyk: Do you really need to separate anything? Instead of calling this method can you just do this:

```
if (args[0] == 'help')
{
   PrintHelp();
   return;
}
// carry on using args
```
and remove this method completely?

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
```diff
@@ -20,7 +20,7 @@ public void RestartProcessOnFileChange()
                 // Wait for the process to start
                 using (var wait = new WaitForFileToChange(scenario.StartedFile))
                 {
-                    scenario.RunDotNetWatch($"{scenario.StatusFile} --no-exit");
+                    scenario.RunDotNetWatch($"run {scenario.StatusFile} --no-exit");
```
moozzyk: What is `--no-exit`? Is it an argument for the test app?

moozzyk:
:ship::it:

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
victorhurdugaci: Yes. That test app can exit after it completes, or not.

victorhurdugaci:
FYI, updated

moozzyk:
LGTM

PR #110, opened 2016-05-25
Make RestartProcessThatTerminatesAfterFileChange more reliable

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/98

I believe the thread sleep will help on Windows too.

please review @moozzyk

### DIFF
--- a/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
+++ b/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Threading;
 using Xunit;
 
 namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
@@ -79,6 +80,11 @@ namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
                 // Then wait for it to restart when we change a file
                 using (var wait = new WaitForFileToChange(scenario.StartedFile))
                 {
+                    // On Unix the file write time is in 1s increments;
+                    // if we don't wait, there's a chance that the polling
+                    // watcher will not detect the change
+                    Thread.Sleep(1000);
+
                     var fileToChange = Path.Combine(scenario.TestAppFolder, "Program.cs");
                     var programCs = File.ReadAllText(fileToChange);
                     File.WriteAllText(fileToChange, programCs);
--- a/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
+++ b/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
@@ -51,7 +51,7 @@ namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
             }
         }
 
-        [Fact(Skip = "aspnet/dotnet-watch#98")]
+        [Fact]
         public void RestartProcessThatTerminatesAfterFileChange()
         {
             using (var scenario = new NoDepsAppScenario())
--- a/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
+++ b/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
@@ -44,6 +44,8 @@ namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
 
         public static void WaitForProcessToStop(int processId, TimeSpan timeout, bool expectedToStop, string errorMessage)
         {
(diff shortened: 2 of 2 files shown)


test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
```diff
@@ -51,7 +52,7 @@ public void RestartProcessOnFileChange()
             }
         }
 
-        [Fact(Skip = "aspnet/dotnet-watch#98")]
+        [Fact]
```
dougbu: Suggest doing this in a separate commit. That way, `git log` can show dc9fead has been reverted.

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
```diff
@@ -44,6 +44,8 @@ public static void WaitForFileToBeReadable(string file, TimeSpan timeout)
 
         public static void WaitForProcessToStop(int processId, TimeSpan timeout, bool expectedToStop, string errorMessage)
         {
+            Console.WriteLine($"Waiting for process {processId} to stop...");
```
dougbu: Remove `WriteLine()`s before checking in. These don't help when the tests are working and just increase the noise level.

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
```diff
@@ -79,6 +80,11 @@ public void RestartProcessThatTerminatesAfterFileChange()
                 // Then wait for it to restart when we change a file
                 using (var wait = new WaitForFileToChange(scenario.StartedFile))
                 {
+                    // On Unix the file write time is in 1s increments;
+                    // if we don't wait, there's a chance that the polling
+                    // watcher will not detect the change
+                    Thread.Sleep(1000);
```
dougbu: A second isn't that long. But is it possible to make this test poll? I'm hoping it can wait at most 1s (or perhaps 5s) but normally complete in a few nanoseconds.

victorhurdugaci: We could probably poll but it would be extra logic for such a small gain. To poll you'd have to go to the disk, get the file, get the file's latest write time, see if the current time is at least 1 second later than that. If not, wait a bit and either do that again or wait 1 second anyway. That's a lot of logic for the hope that you can save at most 1 second in a test run that takes 3-4 minutes

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
victorhurdugaci: That line is very useful, otherwise you see the test doing nothing for a few seconds and then it fails. Then you don't know if it failed because the timeout was wrong or because it really timedout

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
victorhurdugaci: Okay. I split this PR in two commits

victorhurdugaci:
Updated to include 2 commits (per @dougbu's request)

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
dougbu: :ok:

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
dougbu: thx

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
dougbu: Since we don't have and likely won't get a guideline about noise from tests, :ok:

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Waiters.cs
```diff
@@ -62,6 +67,7 @@ public static void WaitForProcessToStop(int processId, TimeSpan timeout, bool ex
             {
                 if (process == null || process.HasExited)
                 {
+                    Console.WriteLine($"Process {processId} no longer running");
```
dougbu: _is_ no longer running? :trollface:

dougbu:
:shipit:

PR #112, opened 2016-05-25
Fix excluded files and folder deletion triggers

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/90, https://github.com/aspnet/dotnet-watch/issues/93

- Bugfix: Trigger on new project.json file structure
- Bugfix: Trigger on folder deletion
- Add test for folder deletion

Please review @moozzyk @kichalla @ajaybhargavb

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.DotNet.ProjectModel.Files;
 using Microsoft.DotNet.ProjectModel.Graph;
 
 namespace Microsoft.DotNet.Watcher.Core.Internal
@@ -15,15 +17,34 @@ namespace Microsoft.DotNet.Watcher.Core.Internal
             ProjectFile = runtimeProject.ProjectFilePath;
             ProjectDirectory = runtimeProject.ProjectDirectory;
 
-            Files = runtimeProject.Files.SourceFiles.Concat(
-                    runtimeProject.Files.ResourceFiles.Values.Concat(
-                    runtimeProject.Files.PreprocessSourceFiles.Concat(
-                    runtimeProject.Files.SharedFiles))).Concat(
-                    new string[] { runtimeProject.ProjectFilePath })
-                .ToList();
+            var compilerOptions = runtimeProject.GetCompilerOptions(targetFramework: null, configurationName: null);
+
+            var filesToWatch = new List<string>() { runtimeProject.ProjectFilePath };
+            if (compilerOptions?.CompileInclude != null)
+            {
+                filesToWatch.AddRange(compilerOptions.CompileInclude.ResolveFiles());
+            }
+            else
+            {
+                filesToWatch.AddRange(runtimeProject.Files.SourceFiles);
+            }
+
+            if (compilerOptions?.EmbedInclude != null)
+            {
+                filesToWatch.AddRange(compilerOptions.EmbedInclude.ResolveFiles());
+            }
+            else
+            {
+                filesToWatch.AddRange(runtimeProject.Files.ResourceFiles.Values);
+            }
+
+            filesToWatch.AddRange(runtimeProject.Files.SharedFiles);
+            filesToWatch.AddRange(runtimeProject.Files.PreprocessSourceFiles);
+
+            Files = filesToWatch;
 
             var projectLockJsonPath = Path.Combine(runtimeProject.ProjectDirectory, "project.lock.json");
-            
+
             if (File.Exists(projectLockJsonPath))
             {
                 var lockFile = LockFileReader.Read(projectLockJsonPath, designTime: false);
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs
@@ -107,6 +107,10 @@ namespace Microsoft.DotNet.Watcher.Core.Internal
                 foreach (var file in project.Files)
                 {
                     closure.Add(file);
+
+                    // We need to add the folder because folder deletions only trigger 
+                    // for the folder, not for the files inside it
+                    closure.Add(Path.GetDirectoryName(file));
                 }
 
                 foreach (var dependency in project.ProjectDependencies)
--- a/src/Microsoft.DotNet.Watcher.Core/IncludeContextExtensions.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/IncludeContextExtensions.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.ProjectModel.Files;
+
+namespace Microsoft.DotNet.Watcher.Core
+{
+    internal static class IncludeContextExtensions
+    {
+        public static IEnumerable<string> ResolveFiles(this IncludeContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            return IncludeFilesResolver
+                .GetIncludeFiles(context, "/", diagnostics: null)
+                .Select(f => f.SourcePath);
+        }
+    }
+}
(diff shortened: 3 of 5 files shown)


src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -15,12 +17,27 @@ public Project(ProjectModel.Project runtimeProject)
...
-                    new string[] { runtimeProject.ProjectFilePath })
-                .ToList();
+            var compilerOptions = runtimeProject.GetCompilerOptions(targetFramework: null, configurationName: null);
+
+            var filesToWatch = new List<string>() { runtimeProject.ProjectFilePath };
+            if (compilerOptions.CompileInclude != null)
+            {
+                filesToWatch.AddRange(compilerOptions.CompileInclude.ResolveFiles());
+
+                if (compilerOptions.EmbedInclude != null)
```
ajaybhargavb: Should this be nested?

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -15,12 +17,27 @@ public Project(ProjectModel.Project runtimeProject)
...
-            Files = runtimeProject.Files.SourceFiles.Concat(
-                    runtimeProject.Files.ResourceFiles.Values.Concat(
-                    runtimeProject.Files.PreprocessSourceFiles.Concat(
-                    runtimeProject.Files.SharedFiles))).Concat(
-                    new string[] { runtimeProject.ProjectFilePath })
-                .ToList();
+            var compilerOptions = runtimeProject.GetCompilerOptions(targetFramework: null, configurationName: null);
+
+            var filesToWatch = new List<string>() { runtimeProject.ProjectFilePath };
+            if (compilerOptions.CompileInclude != null)
```
kichalla: I don't know about the object model here, but is `CompileInclude` the final list of all files that need to be compiled? (i.e considering Include, IncludeFiles,Exclude,ExcludeFiles etc.)

src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs
```diff
@@ -107,6 +107,10 @@ private void GetProjectFilesClosure(string projectFile, ISet<string> closure)
                 foreach (var file in project.Files)
                 {
                     closure.Add(file);
+
+                    // We need to add the folder because folder deletions only trigger 
+                    // for the folder, not for the files inside it
+                    closure.Add(Path.GetDirectoryName(file));
```
kichalla: Have you checked if watching the folder causes any triggers to be fired for any changes to the files underneath it, including the excluded files?

kichalla:
:watch:

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -15,12 +17,27 @@ public Project(ProjectModel.Project runtimeProject)
             ProjectFile = runtimeProject.ProjectFilePath;
             ProjectDirectory = runtimeProject.ProjectDirectory;
 
-            Files = runtimeProject.Files.SourceFiles.Concat(
-                    runtimeProject.Files.ResourceFiles.Values.Concat(
-                    runtimeProject.Files.PreprocessSourceFiles.Concat(
-                    runtimeProject.Files.SharedFiles))).Concat(
-                    new string[] { runtimeProject.ProjectFilePath })
-                .ToList();
+            var compilerOptions = runtimeProject.GetCompilerOptions(targetFramework: null, configurationName: null);
```
moozzyk: Can you have files included/excluded per configuration/framework?

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
moozzyk: Can compilerOptions was ever null?

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -15,12 +17,27 @@ public Project(ProjectModel.Project runtimeProject)
...
+            if (compilerOptions.CompileInclude != null)
+            {
+                filesToWatch.AddRange(compilerOptions.CompileInclude.ResolveFiles());
+
+                if (compilerOptions.EmbedInclude != null)
+                {
+                    filesToWatch.AddRange(compilerOptions.EmbedInclude.ResolveFiles());
+                }
+            }
+            else
```
moozzyk: should this be additive?

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
ajaybhargavb: @kichalla, yes.
@moozzyk, no.

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
victorhurdugaci: That's a good question, maybe you have an answer. So this new project model has two sets of properties: the old ones and the new ones. 

1. Can you have both of them set at the same time? 
1. Does having one set mean ignore the other one? 
1. Which one takes priority because I see that the new properties are `null` by default if not set but the old ones have default values if not set

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
victorhurdugaci: See my questions for @ajaybhargavb

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
victorhurdugaci: @ajaybhargavb this is a question for you

src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs
victorhurdugaci: Oh, that's a good point. Let me check

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
ajaybhargavb: > Can you have both of them set at the same time? Does having one set mean ignore the other one?

Yes. The new ones will take priority.

> Which one takes priority because I see that the new properties are null by default if not set but the old ones have default values if not set

When nothing is set, the default of old schema is used.

ajaybhargavb: In this case, if `CompileInclude` is not set but `EmbedInclude` is set it will be ignored.
This should probably be refactored to something like this.
```c#
if (compilerOptions.CompileInclude != null)
{
    filesToWatch.AddRange(compilerOptions.CompileInclude.ResolveFiles());
}
else
{
    filesToWatch.AddRange(runtimeProject.Files.SourceFiles);
}

if (compilerOptions.EmbedInclude != null)
{
    filesToWatch.AddRange(compilerOptions.EmbedInclude.ResolveFiles());
}
else
{
    filesToWatch.AddRange(runtimeProject.Files.ResourceFiles.Values);
}

filesToWatch.AddRange(runtimeProject.Files.PreprocessSourceFiles);
filesToWatch.AddRange(runtimeProject.Files.SharedFiles);
```

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
ajaybhargavb: `compile` and `embed` are part of `buildOptions` which can be framework/configuration specific. So yes the schema supports it. But I am not sure if it makes sense to include/exclude files for specific framework/configuration.

moozzyk: I think there are rare scenarios like that but I am not sure if we need to proactively address them now.

src/Microsoft.DotNet.Watcher.Core/Internal/ProjectWatcher.cs
victorhurdugaci: Seems to work

victorhurdugaci:
Updated

kichalla:
:shipit: from me

src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -15,15 +17,34 @@ public Project(ProjectModel.Project runtimeProject)
...
-            Files = runtimeProject.Files.SourceFiles.Concat(
-                    runtimeProject.Files.ResourceFiles.Values.Concat(
-                    runtimeProject.Files.PreprocessSourceFiles.Concat(
-                    runtimeProject.Files.SharedFiles))).Concat(
-                    new string[] { runtimeProject.ProjectFilePath })
-                .ToList();
+            var compilerOptions = runtimeProject.GetCompilerOptions(targetFramework: null, configurationName: null);
+
+            var filesToWatch = new List<string>() { runtimeProject.ProjectFilePath };
+            if (compilerOptions?.CompileInclude != null)
```
ajaybhargavb: Sorry if my reply was confusing before, I meant that `compilerOptions` will never be null.

ajaybhargavb:
:shipit:

PR #122, opened 2016-06-01
Fix a timing issue in the polling watcher

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/121

cc @BrennanConroy @pranavkm

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
@@ -117,13 +117,20 @@ namespace Microsoft.DotNet.Watcher.Core.Internal
                 {
                     var fileMeta = _knownEntities[fullFilePath];
 
-                    if (fileMeta.FileInfo.LastWriteTime != f.LastWriteTime)
+                    try
                     {
-                        // File changed
-                        RecordChange(f);
-                    }
+                        if (fileMeta.FileInfo.LastWriteTime != f.LastWriteTime)
+                        {
+                            // File changed
+                            RecordChange(f);
+                        }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Watcher.Core/Internal/FileWatcher/PollingFileWatcher.cs
```diff
@@ -117,13 +117,20 @@ private void CheckForChangedFiles()
...
+                        if (fileMeta.FileInfo.LastWriteTime != f.LastWriteTime)
+                        {
+                            // File changed
+                            RecordChange(f);
+                        }
 
-                    _knownEntities[fullFilePath] = new FileMeta(fileMeta.FileInfo, true);
+                        _knownEntities[fullFilePath] = new FileMeta(fileMeta.FileInfo, true);
+                    }
+                    catch(FileNotFoundException)
```
pranavkm: Couldn't you check for file existence instead using try-catch for control flow?

victorhurdugaci: No because by the time the check is done, the file might be gone. There's a window there between two instructions when the file can disappear.

pranavkm: ok. Should we check for file existence prior to checking `LastWriteTime`. Also formatting (need a space after `catch`)

victorhurdugaci: Hm... I think were's okay with just the try-catch because `ForeachEntityInDirectory` calls `dirInfo.EnumerateFileSystemInfos("*.*");` which returns files that exist at the time of the invocation.

pranavkm:
:shipit:

PR #123, opened 2016-06-03
Improve test stability by disposing correctly

victorhurdugaci:
There was a subtle bug in the tests: the dotnet file watcher can fire multiple times for the same item. Therefore, the handler handling change events can run multiple times. There was a timing issue when the first handler would run and complete and the test would proceed with the execution, disposing objects. Then, the handler would run the second time and try to call a method on a disposed object.

The fix is to stop the watcher from raising events after we got what we needed.

Please review: @BrennanConroy @smitpatel @pranavkm 

Approve for askmode: @muratg @Eilon


smitpatel:
LGTM.

BrennanConroy:
:shipit:

smitpatel:
Ping! CI is blocked on this.

victorhurdugaci:
Waiting on @Eilon / @muratg to approve this

PR #138, opened 2016-07-14
Fix null ref exception when referencing csproj

victorhurdugaci:
Fixes https://github.com/aspnet/dotnet-watch/issues/136

The issue is that for xproj files we don't have a project path even though they appear as project references. The fix is to ignore those projects. The consequence is that dotnet watch will not restart the app when code files change in csproj files.

Please review @BrennanConroy

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
+++ b/src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
@@ -48,7 +48,10 @@ namespace Microsoft.DotNet.Watcher.Core.Internal
             if (File.Exists(projectLockJsonPath))
             {
                 var lockFile = LockFileReader.Read(projectLockJsonPath, designTime: false);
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Watcher.Core/Internal/Implementation/Project.cs
```diff
@@ -48,7 +48,11 @@ public Project(ProjectModel.Project runtimeProject)
             if (File.Exists(projectLockJsonPath))
             {
                 var lockFile = LockFileReader.Read(projectLockJsonPath, designTime: false);
-                ProjectDependencies = lockFile.ProjectLibraries.Select(dep => GetProjectRelativeFullPath(dep.Path)).ToList();
+                ProjectDependencies = lockFile.ProjectLibraries
```
moozzyk: do you need to select twice?

BrennanConroy:
Is this a feature we will be enabling in the future?

victorhurdugaci:
Nope. It's not possible. For an xproj, a csproj is just a dll. It has no notion of where it came from

BrennanConroy:
:shipit:

PR #140, opened 2016-07-15
Null ref fix + more logging

victorhurdugaci:
Fixes: https://github.com/aspnet/dotnet-watch/issues/139

- Fix a NullRef exception that was happening when projects have resx files
- Print the changed file name in all cases

Please review @BrennanConroy @moozzyk


BrennanConroy:
:sheep: :it:

PR #153, opened 2016-08-26
Don't use DOTNET_BUILD_VERSION in code.

pranavkm:
Clean up test projects

### DIFF
--- a/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Scenario/ProjectToolScenario.cs
+++ b/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Scenario/ProjectToolScenario.cs
@@ -1,0 +1,178 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Xml.Linq;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
+{
+    public class ProjectToolScenario : IDisposable
+    {
+        private const string NugetConfigFileName = "NuGet.config";
+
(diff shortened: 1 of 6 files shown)


test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Scenario/ProjectToolScenario.cs
```diff
@@ -0,0 +1,178 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
victorhurdugaci: Why did you add this class?

pranavkm: We needed to change the `version` portion of  `AddToolToProject`. It didn't look like this type was used in any other project so it makes sense to localize it.

victorhurdugaci: Just change the other code and make the version a parameter. The code is in the testing repo.

pranavkm: Why is it in a separate repo?

victorhurdugaci: Because it should be used for the iis tool tests.

pranavkm: IIS tools don't have functional tests as yet. Do you plan on using it there? Additionally, the semantics there is pretty different - you're primarily publishing the application and watching the side effects as opposed to running the tool as standalone. It's pretty unlikely a lot of this is re-usable in that context.

pranavkm: Bump.

victorhurdugaci:
:shipit:

PR #159, opened 2016-09-06
Add support for '--' argument separator

natemcmaster:
~~Depends on https://github.com/aspnet/Common/pull/145.~~ ✅ merged

Lays the foundation for allowing dotnet-watch to have command line flags such as `--verbose`.

Update README.md.

cc @victorhurdugaci

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.DotNet.Watcher.Tools
+{
+    internal class CommandLineOptions
+    {
+        public bool IsHelp { get; private set; }
+        public IList<string> RemainingArguments { get; private set; }
+        public static CommandLineOptions Parse(string[] args, TextWriter consoleOutput)
+        {
+            if (args == null)
+            {
+                throw new ArgumentNullException(nameof(args));
+            }
+
+            var app = new CommandLineApplication(throwOnUnexpectedArg: false)
+            {
+                Name = "dotnet watch",
+                FullName = "Microsoft DotNet File Watcher",
+                Out = consoleOutput,
+                AllowArgumentSeparator = true
+            };
+
+            app.HelpOption("-?|-h|--help");
+
+            app.OnExecute(() =>
+            {
+                if (app.RemainingArguments.Count == 0)
+                {
+                    app.ShowHelp();
+                }
+
+                return 0;
+            });
+
+            if (app.Execute(args) != 0)
+            {
+                return null;
+            }
+
+            return new CommandLineOptions
+            {
+                RemainingArguments = app.RemainingArguments,
+                IsHelp = app.IsShowingInformation
+            };
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Watcher.Tools/Program.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/Program.cs
@@ -6,7 +6,6 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.Watcher.Core;
-using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.DotNet.Watcher.Tools
@@ -14,9 +13,24 @@ namespace Microsoft.DotNet.Watcher.Tools
     public class Program
     {
         private readonly ILoggerFactory _loggerFactory;
+        private readonly CancellationToken _cancellationToken;
+        private readonly TextWriter _out;
 
-        public Program()
+        public Program(TextWriter consoleOutput, CancellationToken cancellationToken)
         {
+            if (consoleOutput == null)
+            {
+                throw new ArgumentNullException(nameof(consoleOutput));
+            }
+
+            if (cancellationToken == null)
+            {
+                throw new ArgumentNullException(nameof(cancellationToken));
+            }
+
+            _cancellationToken = cancellationToken;
+            _out = consoleOutput;
+
             _loggerFactory = new LoggerFactory();
 
             var logVar = Environment.GetEnvironmentVariable("DOTNET_WATCH_LOG_LEVEL");
@@ -44,49 +58,44 @@ namespace Microsoft.DotNet.Watcher.Tools
                     ev.Cancel = false;
                 };
 
-                return new Program().MainInternal(args, ctrlCTokenSource.Token);
+                int exitCode;
+                try
+                {
+                    exitCode = new Program(Console.Out, ctrlCTokenSource.Token)
+                        .MainInternalAsync(args)
+                        .GetAwaiter()
+                        .GetResult();
+                }
+                catch (TaskCanceledException)
+                {
+                    // swallow when only exception is the CTRL+C exit cancellation task
+                    exitCode = 0;
+                }
+                return exitCode;
             }
         }
 
-        private int MainInternal(string[] args, CancellationToken cancellationToken)
+        private async Task<int> MainInternalAsync(string[] args)
         {
-            var app = new CommandLineApplication();
-            app.Name = "dotnet-watch";
-            app.FullName = "Microsoft dotnet File Watcher";
-
-            app.HelpOption("-?|-h|--help");
-
-            app.OnExecute(() =>
+            var options = CommandLineOptions.Parse(args, _out);
+            if (options == null)
             {
-                var projectToWatch = Path.Combine(Directory.GetCurrentDirectory(), ProjectModel.Project.FileName);
-                var watcher = DotNetWatcher.CreateDefault(_loggerFactory);
-
-                try
-                {
-                    watcher.WatchAsync(projectToWatch, args, cancellationToken).Wait();
-                }
-                catch (AggregateException ex)
-                {
-                    if (ex.InnerExceptions.Count != 1 || !(ex.InnerException is TaskCanceledException))
-                    {
-                        throw;
-                    }
-                }
-
-                return 0;
-            });
-
-            if (args == null ||
-                args.Length == 0 ||
-                args[0].Equals("--help", StringComparison.OrdinalIgnoreCase) ||
-                args[0].Equals("-h", StringComparison.OrdinalIgnoreCase) ||
-                args[0].Equals("-?", StringComparison.OrdinalIgnoreCase))
-            {
-                app.ShowHelp();
+                // invalid args syntax
                 return 1;
             }
 
-            return app.Execute();
+            if (options.IsHelp)
+            {
+                return 2;
+            }
+
+            var projectToWatch = Path.Combine(Directory.GetCurrentDirectory(), ProjectModel.Project.FileName);
+
+            await DotNetWatcher
+                    .CreateDefault(_loggerFactory)
+                    .WatchAsync(projectToWatch, options.RemainingArguments, _cancellationToken);
+
+            return 0;
         }
     }
 }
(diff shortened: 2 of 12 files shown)


src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -50,43 +69,31 @@ public static int Main(string[] args)
...
         {
-            var app = new CommandLineApplication();
-            app.Name = "dotnet-watch";
-            app.FullName = "Microsoft dotnet File Watcher";
-
-            app.HelpOption("-?|-h|--help");
+            var options = CommandLineOptions.Parse(args, _out);
+            if (options == null)
+            {
+                _error.WriteLine("Unexpected error parsing command line arguments");
```
victorhurdugaci: Does it print details? If not, how do I know the error so I can fix it?

victorhurdugaci:
Is there any plan to add any arguments right now? If not, I'm not sure that we want this change yet /cc @glennc

src/Microsoft.DotNet.Watcher.Tools/Program.cs
natemcmaster: This is a "should never happen" condition...maybe I should use Debug.Assert instead...?

natemcmaster:
I was thinking of adding `--log <LEVEL>` and/or `--verbose`

victorhurdugaci:
Sync with @glennc , I think he should approve that. Right now you have way to pass it. I think if the number of options we plan to add to dotnet watch is small (<5), we shouldn't complicate the command line, especially if those options are used by very few people.

natemcmaster:
Most usages don't need '--'. It's an optional arg that makes definitive which args are "mine" vs "theirs". 

IMO it's not a matter of complicating/simplifying the command line. This is about env variables. Long environment var names are hard to discover and clunky to set. Adding '--' would allows us to do command line args for these settings (e.g. `--log` or `--polling`). 

cc @glennc

natemcmaster:
Talked with @glennc. The clarify this, the primary purpose of '--' is to make deterministic the split between which args belong to dotnet-watch (subset 1) and which should be passed through to the inside command (subset 2).

The only good example we have of this at the moment is '--help'

Input  | How dotnet-watch splits args into the sub command
------------|-----------------------
`dotnet watch --help` | `dotnet watch --help ` / `(no sub args)`
`dotnet watch -- --help` | `dotnet watch` / `--help`

But let's say we start adding more options, like `--verbose`, here are implications:

Input  | How dotnet-watch splits args into the sub command
------------|-----------------------
`dotnet watch --verbose -- run --verbose -- a b c` | `dotnet watch --verbose` / `run --verbose -- a b c`
`dotnet watch -- run --verbose -- a b c` | `dotnet watch --verbose` / `run --verbose -- a b c`
`dotnet watch run --verbose -- a b c` | `dotnet watch` / `run --verbose -- a b c`
`dotnet watch --verbose run` | `dotnet watch --verbose` / `run`
`dotnet watch -- --verbose run` | `dotnet watch` / `--verbose run`
`dotnet watch --verbose --verbose run` | Error!
`dotnet watch --verbose -- --verbose run` | `dotnet watch --verbose` / `--verbose run`

README.md
```diff
@@ -2,29 +2,34 @@ dotnet-watch
...
+
+[![NuGet stable build](https://img.shields.io/nuget/v/Microsoft.DotNet.Watcher.Tools.svg?style=flat-square&label=stable)](https://www.nuget.org/packages/Microsoft.DotNet.Watcher.Tools/)
+[![MyGet nightly build](https://img.shields.io/dotnet.myget/aspnetcore-dev/vpre/Microsoft.DotNet.Watcher.Tools.svg?style=flat-square&label=nightly)](https://dotnet.myget.org/feed/aspnetcore-dev/package/nuget/Microsoft.DotNet.Watcher.Tools)
+
 ### How To Install
 
 Add `Microsoft.DotNet.Watcher.Tools` to the `tools` section of your `project.json` file:
 
-```
+```js
```
campersau: `json`?

natemcmaster: The `json` highlighter doesn't parse the '...' well.

natemcmaster:
Ping @glennc. Are you okay with "--" support in syntax for dotnet-watch?

natemcmaster:
Talked with @glennc. We're okay moving ahead with this.
cc @moozzyk

test/Microsoft.DotNet.Watcher.Tools.Tests/CommandLineOptionsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        [Theory]
+        [InlineData(new object[] { new[] { "-h" } })]
+        [InlineData(new object[] { new[] { "-?" } })]
+        [InlineData(new object[] { new[] { "--help" } })]
+        [InlineData(new object[] { new[] { "--" } })]
+        [InlineData(new object[] { new string[0] })]
+        public void HelpArgs(string[] args)
+        {
+            // arrange
```
moozzyk: nit: I am not sure what is the style in this repo but I personally don't find these comments useful.

src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -44,49 +65,39 @@ public static int Main(string[] args)
...
+            try
             {
-                app.ShowHelp();
-                return 1;
+                DotNetWatcher
+                    .CreateDefault(_loggerFactory)
+                    .WatchAsync(projectToWatch, options.RemainingArguments, _cancellationToken)
+                    .Wait();
+            }
+            catch (AggregateException ex) when (ex.InnerExceptions.Count == 1 && ex.InnerException is TaskCanceledException)
```
moozzyk: you could make MainInternal async and then catch only TaskCanceledException. You would `.Wait()` in Main instead of here.

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            if (args == null)
+            {
+                throw new ArgumentNullException(nameof(args));
+            }
+
+            var app = new CommandLineApplication(throwOnUnexpectedArg: false)
+            {
+                Name = "dotnet watch",
+                FullName = "Microsoft DotNet File Watcher",
```
moozzyk: Is this correct casing (DotNet)?

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        public bool IsHelp { get; set; }
+        public IList<string> RemainingArguments { get; set; }
+        public static CommandLineOptions Parse(string[] args, TextWriter consoleOutput)
+        {
+            if (args == null)
+            {
+                throw new ArgumentNullException(nameof(args));
+            }
+
+            var app = new CommandLineApplication(throwOnUnexpectedArg: false)
```
moozzyk: what happens in an unexpected argument appears before `--`? (e.g. `dotnet watch --help --bogus`)

moozzyk: As per https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/CommandLine/CommandLineApplication.cs#L16-L18 I guess `--bogus` will be added to RemainingArgs and then we will just exit In Program l. 82

test/Microsoft.DotNet.Watcher.Tools.Tests/CommandLineOptionsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using Xunit;
+
+namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
+{
+    public class CommandLineOptionsTests
+    {
+        [Theory]
+        [InlineData(new object[] { new[] { "-h" } })]
+        [InlineData(new object[] { new[] { "-?" } })]
+        [InlineData(new object[] { new[] { "--help" } })]
```
moozzyk: add a case where there is an argument after `--help`

src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -6,17 +6,38 @@
...
-using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.DotNet.Watcher.Tools
 {
     public class Program
     {
         private readonly ILoggerFactory _loggerFactory;
+        private readonly CancellationToken _cancellationToken;
+        private readonly TextWriter _error;
```
moozzyk: Is this being used at all?

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.DotNet.Watcher.Tools
+{
+    internal class CommandLineOptions
+    {
+        public bool IsHelp { get; set; }
```
moozzyk: This should not be settable outside this class.

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        public IList<string> RemainingArguments { get; set; }
```
moozzyk: Not that this matter as much (you can always change contents) but it should not be settable outside this class.

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                Name = "dotnet watch",
+                FullName = "Microsoft DotNet File Watcher",
+                Out = consoleOutput,
+                AllowArgumentSeparator = true
+            };
+
+            app.HelpOption("-?|-h|--help");
+
+            var options = new CommandLineOptions();
```
moozzyk: nit: move down where you actually need it.

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
moozzyk: actually both could have just getters...

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
natemcmaster: Yes. It looks a little weird, but it's what the [DotNet team uses](https://github.com/dotnet/cli/tree/rel/1.0.0/src).

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
natemcmaster: > --bogus will be added to RemainingArgs

Yes, that's what will happen. The heuristic for RemainingArgs is: RemainingArgs = all args starting with the first unrecognized arg or the first arg after the '--'

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
natemcmaster: IReadOnlyList with a private setter?

src/Microsoft.DotNet.Watcher.Tools/Program.cs
natemcmaster: Not at the moment. Will remove.

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
moozzyk: I would stick just to `IList<string>`  with a private setter. No reason to make it more complicated than this...

natemcmaster:
:up: :date:

src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -6,17 +6,31 @@
...
 namespace Microsoft.DotNet.Watcher.Tools
 {
     public class Program
     {
         private readonly ILoggerFactory _loggerFactory;
+        private readonly CancellationToken _cancellationToken;
+        private readonly TextWriter _out;
 
-        public Program()
+        public Program(CancellationToken cancellationToken)
```
moozzyk: Is this ctor needed at all? Why not make the other one public and remove this one? We don't seem to use it (unless there are some tests using it)..

moozzyk:
One small comment and then :ship::it:

src/Microsoft.DotNet.Watcher.Tools/Program.cs
natemcmaster: Yeah, can be removed. Opened this PR so long ago I can't remember what I was thinking when I added the public/internal ctor split.

PR #164, opened 2016-09-19
Add dotnet-sql-cache

natemcmaster:
Moving day! Incoming content from https://github.com/aspnet/Caching/pull/232

cc @Eilon @muratg


Eilon:
cc @kichalla

PR #173, opened 2016-09-20
Minor usability improvements to dotnet-user-secrets

### LINKED ISSUE #170: `dotnet user-secrets {command} --project {path}` does not support relative paths
_From @dougbu on May 12, 2016 17:13_

1. It isn't clear why we need the `--project` option at all. The `dotnet user-secrets` command must be executed in a project directory with `Microsoft.Extensions.SecretManager.Tools` fully configured.
2. If this option is needed, it should support relative paths. That would at least be useful in corner cases where user secrets are used in a group of projects but only one has the tool configured.
3. `dotnet user-secrets list --project $pwd/../../src/UserSecrets.work` is ugly as sin.
4. What is the parameter named `root`?

For example
```
dotnet user-secrets list --project ../../src/UserSecrets.work
```

Results in
```
critical: Command failed : The path must be absolute.
Parameter name: root
```

_Copied from original issue: aspnet/UserSecrets#77_

### LINKED ISSUE #171: Improve `dotnet user-secrets {command} --project {path}` error messages
_From @dougbu on May 12, 2016 17:16_

The error message displayed when `{path}` is invalid is not descriptive and references an argument rather than the command-line option.

For example,
```
dotnet user-secrets list --project \garbage
```
Results in
```
critical: Command failed : The directory name C:\garbage\ is invalid.
Parameter name: path
```

Note the parameter name mentioned above is inconsistent w/ the in #77. It's the same command-line option and the parameter should also be the same.

_Copied from original issue: aspnet/UserSecrets#78_

natemcmaster:
Resolves #170.
Resolves #171.

cc @dougbu

### DIFF
--- a/test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
+++ b/test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.Configuration.UserSecrets.Tests;
 using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Abstractions;
+using Microsoft.Extensions.SecretManager.Tools.Internal;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Tests
 {
@@ -20,6 +21,17 @@ namespace Microsoft.Extensions.SecretManager.Tools.Tests
         public SecretManagerTests(ITestOutputHelper output)
         {
             _logger = new TestLogger(output);
+        }
+
+        [Fact]
+        public void Error_Project_DoesNotExist()
+        {
+            var projectPath = Path.Combine(Path.GetTempPath(), "dne", Guid.NewGuid().ToString(), "project.json");
+            var secretManager = new Program(Console.Out, Directory.GetCurrentDirectory()) { Logger = _logger };
+
+            var ex = Assert.Throws<GracefulException>(() => secretManager.RunInternal("list", "--project", projectPath));
+
+            Assert.Equal(ex.Message, Resources.FormatError_ProjectPath_NotFound(projectPath));
         }
 
         [Theory]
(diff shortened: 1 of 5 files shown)


natemcmaster:
cc @mikaelm12 for review

test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
```diff
@@ -10,18 +10,60 @@
...
+
+        [Fact]
+        public void Error_Project_DoesNotExist()
+        {
+            var projectPath = Path.Combine(GetTempSecretProject(), "does_not_exist", "project.json");
+            var secretManager = new Program(Console.Out, Directory.GetCurrentDirectory()) { Logger = _logger };
+
+            var ex = Assert.Throws<GracefulException>(() => secretManager.RunInternal("list", "--project", projectPath));
+
+            Assert.Equal(ex.Message, Resources.FormatError_ProjectPath_NotFound(projectPath));
```
mikaelm12: Nit: Shouldn't the order of these arguments be expected then actual

test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
```diff
@@ -35,7 +77,7 @@ public void SetSecrets(bool fromCurrentDirectory)
                             new KeyValuePair<string, string>("key2", string.Empty)
                         };
 
-            var projectPath = UserSecretHelper.GetTempSecretProject();
+            var projectPath = GetTempSecretProject();
```
mikaelm12: Extra space?

test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
natemcmaster: Good catch. Yes, it is the other way around.

test/Microsoft.Extensions.SecretManager.Tools.Tests/SecretManagerTests.cs
natemcmaster: Can you clarify...do you mean I should add a space?

mikaelm12: Sorry, I meant is the empty line 81 nedded

natemcmaster: No, can be removed.

mikaelm12:
🚢 it

PR #177, opened 2016-09-26
Remove Microsoft.DotNet.Watcher.Core package

natemcmaster:
The shared watch code isn't an API we really want to support. This removes the Microsoft.DotNet.Watcher.Core and merges code into once single ".Tools" project.

History: in the days of DNX, there was a plan to share code between dnx-watch and Visual Studio tooling to watch the system. We split the package into two pieces with this intention. Since then, dotnet-cli has replaced DNX and MSBuild will replace project.json. 

cc @muratg @victorhurdugaci nothing in aspnet/Universe is using this package. Is there anyone we need to notify about this removal?


muratg:
I like it, but isn't it a breaking change for 1.x?

natemcmaster:
We never RTMed the package. It was only a "preview" so I think we are clear.

victorhurdugaci:
As long as you don't remove the old packages, it shouldn't break.

natemcmaster:
> As long as you don't remove the old packages, it shouldn't break.

Yeah, we'll leave it on nuget.org. @muratg do we plan to unlist deprecated, prerelease packages from nuget.org?

muratg:
@natemcmaster There's a longer term plan to unlist the older packages on nuget.org. We'll eventually do this, up to RC2 for now, I think.

PR #178, opened 2016-09-28
Create initial prototype of dotnet-user-secrets with MSBuild support

natemcmaster:
A prototype port of "dotnet-user-secrets" from project.json to MSBuild support.

This CLI project tool requires the ability to evaluate the project file to find the 'UserSecretsId' property. At the moment, this change will introduce a dependency on the existence and location of the .NET Core SDK. Without this, we cannot properly set the "MSBuildExtensionsPath" property which leads to project evaluation failing on initialization.

NB: merging to feature/msbuild, not dev. More refinement needed before this is ready for a nightly version.

TODO:

- [x] Can we remove dependency on existence of MSBuild inside .NET Core SDK? (I hope so.) _Not for now. Maybe for 1.0.0 we can._
- [x] How to publish this thing as a nupkg? (see https://github.com/Microsoft/msbuild/issues/1097 and https://github.com/Microsoft/msbuild/issues/1096) _Punting this for now. Will address when CLI stabilizes._
- [x] Better error handling project evaluation fails to find SDK
- [x] Provide upgrade path from project.json to msbuild - see https://github.com/dotnet/cli/pull/4326
- [x] Update Microsoft.Extensions.Configuration.UserSecrets to support MSBuild scenarios - https://github.com/aspnet/Configuration/pull/525

cc @NTaylorMullen @bricelam @prafullbhosale @davidfowl @jeffkl @eerhardt


src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            var projectDir = Path.GetDirectoryName(fileInfo.PhysicalPath);
+
+            var dotNetHome = Path.GetDirectoryName(new Muxer().MuxerPath);
+            var sdkResolver = new DotNetCoreSdkResolver(dotNetHome);
+            var sdk = sdkResolver.ResolveProjectSdk(projectDir);
+
+            // workaround limitations in using MSBuild API
+            Environment.SetEnvironmentVariable("MSBUILD_EXE_PATH", Path.Combine(sdk.BasePath, "MSBuild.exe"));
```
jeffkl: This is interesting it looks like you're reusing the MSBuild that ships with the dotnet CLI is that correct?

natemcmaster: Yes, that's the only way I've been able to get this working.... :-/

jeffkl: Is this running as one of these "project tools" where it's not deployed via `dotnet publish` then?

Or do you get errors about missing `.props` or `.targets` files?

natemcmaster: Yeah, this one of those "project tools". It's deployed into a user project by dotnet-cli with that funky ".nuget/packages/.tools/" folder magic. 

The error is https://github.com/Microsoft/msbuild/issues/1097. I've also attempted to use a version of MSBuild that is shipped with my tool, but have found issues getting all of MSBuild dependencies into one place (https://github.com/Microsoft/msbuild/issues/1096)...which is why for now this prototype finds the MSBuild inside the .NET Core SDK.

jeffkl: Okay yeah we don't have a good story yet for these project tools.  I guess this is valid but hopefully this isn't abused.

I'm also working on a change to build `MSBuild.dll` instead of `MSBuild.exe` like we should be.  This would break this tool if you try and take a newer package and I'm still investigating who else will be broken.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
```diff
@@ -0,0 +1,63 @@
...
+                var globalProperties = new Dictionary<string, string>
+                {
+                    { "DesignTimeBuild", "true" },
+                    { "MSBuildExtensionsPath", sdk.BasePath }
+                };
+
+                project = new Project(xml, globalProperties, /* toolsVersion: */ null, projectCollection);
+            }
+
+            var userSecretsId = project.Properties.FirstOrDefault(p => p.Name == "UserSecretsId")?.EvaluatedValue;
```
jeffkl: Is there a reason you're not using `GetPropertyValue`?
https://msdn.microsoft.com/en-us/library/microsoft.build.evaluation.project.getpropertyvalue.aspx

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.ProjectModel.Internal;
+using GracefulException = Microsoft.Extensions.SecretManager.Tools.Internal.GracefulException;
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal
+{
+    internal class MsBuildProjectReader
+    {
+        public string ReadUserSecretId(IFileInfo fileInfo)
```
jeffkl: If I wrote this class, I would have it's constructor take `IFileInfo`, add a private method like `GetPropertyValue(string name)`, and change `ReadUserSecretId()` to call it passing `"UserSecretsId"`.  This way you could add new methods to read new properties.  But I have no idea the future of this tool so do what you want.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
natemcmaster: I assumed it would throw if the property didn't exist, but the API doc says otherwise, so I'll update to use that API.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
natemcmaster: Yeah...i'm kind of intentionally trying to make this class useless to others at the moment. I don't want others to copy-paste without understanding its implementation....plus I'm not sure the implementation is good...still learning which MSBuild APIs are appropriate to use.

But once I'm more confident in my MSBuild-fu, I'll make changes as you suggested and will likely introduce a implementation that can be shared with all aspnet tools that need to evaluate projects.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
natemcmaster: > I guess this is valid but hopefully this isn't abused.

IIUC this is exists primarily as a test hook. I'm kind of hoping Microsoft/msbuild#1097 results in some API that we can use to tell MSBuild "here's the runtime you should use".

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdk.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.ProjectModel.Internal
+{
+    internal class DotNetCoreSdk
```
troydai: as far as i remember, we keep the types in `*.Internal` public. it is to avoid over-internalized types but explicitly states that types in `*.Internal` namespace are subject to change.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+
+namespace Microsoft.Extensions.ProjectModel.Internal
+{
+    internal class DotNetCoreSdkResolver
```
troydai: same as above.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private readonly string _installationPath;
+
+        public DotNetCoreSdkResolver(string installationDir)
+        {
+            _installationPath = installationDir;
```
troydai: why not keep the names consistent?

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+            return new DotNetCoreSdk
+            {
+                BasePath = first.path,
+                Version = first.version.ToFullString()
+            };
+        }
+
+        public DotNetCoreSdk ResolveProjectSdk(string projectDir)
+        {
+            var global = ResolveGlobalJsonSdkVersion(projectDir);
```
troydai: use a better variable name: `sdkVersion`

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public DotNetCoreSdkResolver(string installationDir)
+        {
+            _installationPath = installationDir;
+        }
+
+        private IEnumerable<string> Installed
+            => Directory.EnumerateDirectories(Path.Combine(_installationPath, "sdk"));
+
+        private string ResolveGlobalJsonSdkVersion(string start)
```
troydai: position private method behind public methods.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+                return obj?.sdk?.version;
+            }
+            catch (JsonException)
+            {
+                // TODO log
+                return null;
+            }
+        }
+
+        public DotNetCoreSdk ResolveLatest()
```
troydai: can this method be private?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -142,52 +140,48 @@ internal int RunInternal(params string[] args)
...
+            else
             {
-                throw new GracefulException(Resources.FormatError_ProjectPath_NotFound(projectPath));
+                fileInfo = new FileInfo(projectPath);
             }
 
-            Logger.LogDebug(Resources.Message_Project_File_Path, fileInfo.PhysicalPath);
-            return ReadUserSecretsId(fileInfo);
-        }
+            var physicalFile = new PhysicalFileInfo(fileInfo);
```
troydai: create this instance after test if fileInfo exists.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
victorhurdugaci: Do you need an `IFileInfo` or could you just use `Stream` instead (and maybe keep this overload too)? It makes testing easier

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
victorhurdugaci: Just a note here in case you plan to use this code in `dotnet watch`. When you spin a new child process, it inherits the environment variables. Not sure if this will impact the watched app but it might have some interesting side effects

src/Microsoft.Extensions.SecretManager.Tools/Internal/CommandLineOptions.cs
```diff
@@ -33,6 +34,11 @@ public static CommandLineOptions Parse(string[] args, TextWriter output)
             var optionProject = app.Option("-p|--project <PROJECT>", "Path to project, default is current directory",
                 CommandOptionType.SingleValue, inherited: true);
 
+            // the escape hatch if project evaluation fails, or if users want to alter a secret store other than the one
+            // in the current project
+            var optionId = app.Option("--id", "The user secret id to use.",
```
NTaylorMullen: When would project evaluation fail? Trying to determine if we really need this.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private readonly string _installationPath;
+
+        public DotNetCoreSdkResolver(string installationDir)
+        {
+            _installationPath = installationDir;
+        }
+
+        private IEnumerable<string> Installed
+            => Directory.EnumerateDirectories(Path.Combine(_installationPath, "sdk"));
```
NTaylorMullen: Given the fact that we're already tying this expectation to an SDK folder there's no value in taking in the `installationDir`. Lets group all of our questionable "where is msbuild.exe" assumptions in this one class.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+                    globalJson = new FileInfo(candidate);
+                    break;
+                }
+                dir = dir.Parent;
+            }
+            if (globalJson == null)
+            {
+                return null;
+            }
+            var t = new { sdk = new { version = "" } };
```
NTaylorMullen: Make a private class, this is super hacky. That will enable you to not declare this here. See next coment

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+            try
+            {
+                var obj = JsonConvert.DeserializeAnonymousType(File.ReadAllText(globalJson.FullName), t);
```
NTaylorMullen: Break this statement apart. Aka:
```C#
var globalJsonContents = File.ReadAlText(globalJson.FullName);
var globalJson =JsonConvert.Deserialize<GlobalJson>(globalJsonContents);
```

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+                // TODO log
+                return null;
+            }
+        }
+
+        public DotNetCoreSdk ResolveLatest()
+        {
+            var first = Installed.Select(d => new { path = d, version = SemanticVersion.Parse(Path.GetFileName(d)) })
+                .OrderByDescending(sdk => sdk.version)
+                .First();
```
NTaylorMullen: Do we want to handle cases when SDK's haven't been installed?

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,89 @@
...
+                    .First();
+
+                return new DotNetCoreSdk
+                {
+                    BasePath = sdk.path,
+                    Version = sdk.version.ToFullString()
+                };
+            }
+
+            return ResolveLatest();
```
NTaylorMullen: Odd that above you're newing up a `DotNetCoreSDK` in the case that you can resolve a globaljson sdk but not in this case. It's inconsistent. 

If you're going to have the ResolveLatest/ResolveGlobalJsonSdkVersion methods separate you should normalize their output.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.ProjectModel.Internal;
+using GracefulException = Microsoft.Extensions.SecretManager.Tools.Internal.GracefulException;
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal
+{
+    internal class MsBuildProjectReader
```
NTaylorMullen: Doesn't look like this class has any state. Any value in it not being static?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -142,52 +140,48 @@ internal int RunInternal(params string[] args)
...
                 projectPath = Path.Combine(_workingDirectory, projectPath);
             }
 
-            if (!projectPath.EndsWith("project.json", StringComparison.OrdinalIgnoreCase))
+            var directoryInfo = new DirectoryInfo(projectPath);
+            FileInfo fileInfo;
+            if (directoryInfo.Exists)
             {
-                projectPath = Path.Combine(projectPath, "project.json");
+                var projects = directoryInfo.EnumerateFileSystemInfos("*.csproj").ToList();
```
NTaylorMullen: We should normalize all this logic across our "complex" tools. Also, your repo where you normalize project.json and MSBuild project reading seem like the correct approach to move forward. Aka, our tools need to continue supporting both project.json and MSBuild for a little while, might as well make the process painless for all 😄

src/Microsoft.Extensions.SecretManager.Tools/Internal/CommandLineOptions.cs
natemcmaster: > When would project evaluation fail?

At the moment, the question is "When does project evaluation **not** fail?" :trollface: 

Bad XML, missing imports, invalid syntax, etc... there are lots of reasons why it's good to have a an escape hatch. Plus, some may prefer to just us `--id` and avoid the MSBuild complexity.

src/Microsoft.Extensions.SecretManager.Tools/Internal/DotNetCoreSdkResolver.cs
natemcmaster: In theory, it's not possible for "dotnet user-secrets" to ever be invoked without some SDK installed...but at the very least we need a better error message for that.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
natemcmaster: I don't think it will be too big an issues as dotnet-watch will shell out to the SDK which takes care of ensuring environment variables are set correctly before MSBuild.exe is invoked.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectReader.cs
natemcmaster: IFileInfo helps also capture info about the directory which we need to ensure we get the right SDK version.

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
natemcmaster: Yeah, I'm all for recreating some kind of "ProjectResolver" that finds either project.json or msbuild files.

natemcmaster:
:up: :date: responded to much of the feedback above. Refactored MSBuild evaluation into reusable components.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectContextFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal
+{
+    public class MsBuildProjectContextFactory
+    {
+        private readonly DotNetCoreSdkResolver _sdkResolver;
+
+        public MsBuildProjectContextFactory()
+        {
+            var dotNetHome = Path.GetDirectoryName(new Muxer().MuxerPath);
```
NTaylorMullen: Why not group this logic in the DotNetCoreSdkResolver?

natemcmaster: In https://github.com/aspnet/DotNetTools/pull/180 I have moved this into an abstraction that represents `MsBuildContext`

natemcmaster:
:up: :date: rebased on top of project model abstractions.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.FileProviders.Physical;
+
+namespace Microsoft.Extensions.SecretManager.Tools
+{
+    internal class MsBuildProjectFinder
```
natemcmaster: Would this be useful in project model abstractions too?

NTaylorMullen: Yes. If you put it into PR i'll review it quickly 😄

natemcmaster: @NTaylorMullen added to https://github.com/aspnet/DotNetTools/pull/181

NTaylorMullen:
Any reason to not support both project.json and msbuild?

natemcmaster:
@NTaylorMullen I would assume we won't want to support both PJ and MSBuild, but I've opened that up for discussion with our PMs.

natemcmaster:
Ping @NTaylorMullen any other changes needed before I merge to <kbd>feature/msbuild</kbd> ?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -134,59 +136,37 @@ internal int RunInternal(params string[] args)
                 CommandOutputProvider.LogLevel = LogLevel.Debug;
             }
 
-            var userSecretsId = ResolveUserSecretsId(options);
+            var userSecretsId = !string.IsNullOrEmpty(options.Id)
```
NTaylorMullen: nit: change order for clarity:

```C#
var userSecretsId = string.IsNullOrEmpty(options.Id) ? ResolveIdFromProject(options.Project) : options.Id;
```

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -134,59 +136,37 @@ internal int RunInternal(params string[] args)
...
-            if (fileInfo == null || !fileInfo.Exists)
+            try
             {
-                throw new GracefulException($"Could not find file '{fileInfo.PhysicalPath}'");
+                var project = new MsBuildProjectContextBuilder()
+                    .UseMsBuild(_msBuildContext)
+                    .AsDesignTimeBuild()
+                    .WithBuildTargets(Array.Empty<string>())
+                    .WithProjectFile(projectFile)
+                    .WithTargetFramework("") // TFM doesn't matter
```
NTaylorMullen: Why do we need to set it then? Seems like this should be a default

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -134,59 +136,37 @@ internal int RunInternal(params string[] args)
...
+                    .Build();
+
+                return project.GetUserSecretsId();
```
NTaylorMullen: Make debugging easier
```C#
var id = project.GetUserSecretsId();
return id;
```

test/Microsoft.Extensions.ProjectModel.Tests/project.json
```diff
@@ -12,10 +12,7 @@
     "Microsoft.DotNet.ProjectModel": "1.0.0-rc3-003121",
     "Microsoft.Build.Runtime": "15.1.298-preview5",
     "Microsoft.Extensions.FileProviders.Physical": "1.1.0-*",
-    "Microsoft.Extensions.ProjectModel.Sources": {
-        "type": "build",
-        "version": "1.0.0-*"
-    },
+    "Microsoft.Extensions.ProjectModel.Sources": "1.0.0-*",
```
NTaylorMullen: Why not build time dependency?

natemcmaster: Less typing (shrugs lazily). It's a test project so doesn't matter either way.

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
natemcmaster: This tool doesn't need to worry about TFMs. We just want to read a global property. The alternative would be `.BuildAllTargetFrameworks().First()`

PR #180, opened 2016-10-03
Create replacement API for Microsoft.DotNet.ProjectModel

natemcmaster:
Creates a new package "Microsoft.Extensions.ProjectModel.Sources" which contains API for project evaluation to replace Microsoft.DotNet.ProjectModel.

Follows patterns from Mircrosoft.DotNet.ProjectModel

cc @NTaylorMullen @bricelam @prafullbhosale @muratg @Eilon

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -21,5 +21,11 @@ k-standard-goals
               "-n src/Microsoft.Extensions.SecretManager.Tools/Microsoft.Extensions.SecretManager.Tools.nuspec " +
               "-n src/Microsoft.Extensions.Caching.SqlConfig.Tools/Microsoft.Extensions.Caching.SqlConfig.Tools.nuspec " +
               "-n src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.nuspec ");
+
+          DotnetPack("src/Microsoft.Extensions.ProjectModel.Sources/project.json", BUILD_DIR_LOCAL, E("Configuration"), E("KOREBUILD_DOTNET_PACK_OPTIONS") + " --no-build");
+          foreach (var nupkg in Files.Include(Path.Combine(BUILD_DIR_LOCAL, "*/" + E("Configuration") + "/*.nupkg")))
+          {
+              File.Copy(nupkg, Path.Combine(BUILD_DIR_LOCAL, Path.GetFileName(nupkg)), true);
+          }
         }
     }
--- a/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
@@ -1,0 +1,102 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.DotNet.ProjectModel;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class DotNetProjectContext : IProjectContext
+    {
+        private readonly ProjectContext _projectContext;
+        private readonly OutputPaths _paths;
+        private readonly Lazy<JObject> _rawProject;
+        private readonly CommonCompilerOptions _compilerOptions;
+
+        public DotNetProjectContext(ProjectContext projectContext, string configuration, string outputPath)
+        {
+            if (projectContext == null)
+            {
+                throw new ArgumentNullException(nameof(projectContext));
+            }
+
+            if (string.IsNullOrEmpty(configuration))
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            _rawProject = new Lazy<JObject>(() =>
+                {
+                    using (var stream = new FileStream(projectContext.ProjectFile.ProjectFilePath, FileMode.Open, FileAccess.Read))
+                    using (var streamReader = new StreamReader(stream))
+                    using (var jsonReader = new JsonTextReader(streamReader))
+                    {
+                        return JObject.Load(jsonReader);
+                    }
+                });
+
+            Configuration = configuration;
+            _projectContext = projectContext;
+
+            _paths = projectContext.GetOutputPaths(configuration, buidBasePath: null, outputPath: outputPath);
+            _compilerOptions = _projectContext.ProjectFile.GetCompilerOptions(TargetFramework, Configuration);
+
+            // Workaround https://github.com/dotnet/cli/issues/3164
+            IsClassLibrary = !(_compilerOptions.EmitEntryPoint
+                    ?? projectContext.ProjectFile.GetCompilerOptions(null, configuration).EmitEntryPoint.GetValueOrDefault());
+        }
+
+        public bool IsClassLibrary { get; }
+
+        public NuGetFramework TargetFramework => _projectContext.TargetFramework;
+        public string Config => _paths.RuntimeFiles.Config;
+        public string DepsJson => _paths.RuntimeFiles.DepsJson;
+        public string RuntimeConfigJson => _paths.RuntimeFiles.RuntimeConfigJson;
+        public string PackagesDirectory => _projectContext.PackagesDirectory;
+
+        public string AssemblyFullPath =>
+            !IsClassLibrary && (_projectContext.IsPortable || TargetFramework.IsDesktop())
+                ? _paths.RuntimeFiles.Executable
+                : _paths.RuntimeFiles.Assembly;
+
+        public string Configuration { get; }
+        public string ProjectFullPath => _projectContext.ProjectFile.ProjectFilePath;
+        public string ProjectName => _projectContext.ProjectFile.Name;
+        // TODO read from xproj if available
+        public string RootNamespace => _projectContext.ProjectFile.Name;
+        public string TargetDirectory => _paths.RuntimeOutputPath;
+        public string Platform => _compilerOptions.Platform;
+
+        public IEnumerable<string> CompilationItems
+            => _compilerOptions.CompileInclude.ResolveFiles();
+
+        public IEnumerable<string> EmbededItems
+            => _compilerOptions.EmbedInclude.ResolveFiles();
+
+        /// <summary>
+        /// Returns string values of top-level keys in the project.json file
+        /// </summary>
+        /// <param name="propertyName"></param>
+        /// <param name="propertyNameComparer"></param>
+        /// <returns></returns>
+        public string FindProperty(string propertyName) => FindProperty<string>(propertyName);
+
+        public TProperty FindProperty<TProperty>(string propertyName)
+        {
+            foreach (var item in _rawProject.Value)
+            {
+                if (item.Key.Equals(propertyName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return item.Value.Value<TProperty>();
+                }
+            }
+
+            return default(TProperty);
+        }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
@@ -1,0 +1,129 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+
+namespace Microsoft.Extensions.ProjectModel.Internal
+{
+    internal class DotNetCoreSdkResolver
+    {
+        private readonly string _installationDir;
+
+        /// <summary>
+        /// Represents a resolver that uses the currently executing <see cref="Muxer"/> to find the .NET Core SDK installation
+        /// </summary>
+        public static readonly DotNetCoreSdkResolver DefaultResolver = new DotNetCoreSdkResolver(Path.GetDirectoryName(new Muxer().MuxerPath));
+
+        /// <summary>
+        /// Instantiates a resolver that locates the SDK
+        /// </summary>
+        /// <param name="installationDir">The directory containing dotnet muxer, aka DOTNET_HOME</param>
+        public DotNetCoreSdkResolver(string installationDir)
+        {
+            _installationDir = installationDir;
+        }
+
+        /// <summary>
+        /// Find the latest SDK installation (uses SemVer 1.0 to determine what is "latest")
+        /// </summary>
+        /// <returns>Path to SDK root directory</returns>
+        public DotNetCoreSdk ResolveLatest()
+        {
+            var latest = FindInstalled()
+                .Select(d => new { path = d, version = SemanticVersion.Parse(Path.GetFileName(d)) })
+                .OrderByDescending(sdk => sdk.version)
+                .FirstOrDefault();
+
+            if (latest == null)
+            {
+                throw CreateSdkNotInstalledException();
+            }
+
+            return new DotNetCoreSdk
+            {
+                BasePath = latest.path,
+                Version = latest.version.ToFullString()
+            };
+        }
+
+        public DotNetCoreSdk ResolveProjectSdk(string projectDir)
+        {
+            var sdkVersion = ResolveGlobalJsonSdkVersion(projectDir);
+            if (string.IsNullOrEmpty(sdkVersion))
+            {
+                return ResolveLatest();
+            }
+
+            var sdk = FindInstalled()
+                .Where(p => Path.GetFileName(p).Equals(sdkVersion, StringComparison.OrdinalIgnoreCase))
+                .Select(d => new { path = d, version = SemanticVersion.Parse(Path.GetFileName(d)) })
+                .FirstOrDefault();
+
+            if (sdk == null)
+            {
+                throw CreateSdkNotInstalledException();
+            }
+
+            return new DotNetCoreSdk
+            {
+                BasePath = sdk.path,
+                Version = sdk.version.ToFullString()
+            };
+        }
+
+        private Exception CreateSdkNotInstalledException()
+        {
+            return new DirectoryNotFoundException($"Could not find an installation of the .NET Core SDK in '{_installationDir}'");
+        }
+
+        private IEnumerable<string> FindInstalled()
+            => Directory.EnumerateDirectories(Path.Combine(_installationDir, "sdk"));
+
+        private string ResolveGlobalJsonSdkVersion(string start)
+        {
+            var dir = new DirectoryInfo(start);
+            FileInfo fileInfo = null;
+            while (dir != null)
+            {
+                var candidate = Path.Combine(dir.FullName, "global.json");
+                if (File.Exists(candidate))
+                {
+                    fileInfo = new FileInfo(candidate);
+                    break;
+                }
+                dir = dir.Parent;
+            }
+            if (fileInfo == null)
+            {
+                return null;
+            }
+            try
+            {
+                var contents = File.ReadAllText(fileInfo.FullName);
+                var globalJson = JsonConvert.DeserializeObject<GlobalJsonStub>(contents);
+                return globalJson?.sdk?.version;
+            }
+            catch (JsonException)
+            {
+                // TODO log
+                return null;
+            }
+        }
+
+        private class GlobalJsonStub
+        {
+            public GlobalJsonSdkStub sdk { get; set; }
+
+            public class GlobalJsonSdkStub
+            {
+                public string version { get; set; }
+            }
+        }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.Extensions.ProjectModel.Internal;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    /// <summary>
+    /// Represents the msbuild context used to parse a project model
+    /// </summary>
+    internal class MsBuildContext
+    {
+        public string MsBuildExecutableFullPath { get; private set; }
+        public string ExtensionsPath { get; private set; }
+
+        public static MsBuildContext FromCurrentDotNetSdk()
+        {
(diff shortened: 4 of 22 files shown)


src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                return projectInstance;
+            }
+
+            var logger = new InMemoryLogger();
+            projectInstance.Build(targets, new[] { logger });
+
+            if (logger.Errors.Count > 0)
+            {
+                throw new InvalidOperationException(logger.Errors[0].Message);
```
prafullbhosale: Why only the first one?

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,124 @@
...
+                    fileInfo = new FileInfo(candidate);
+                    break;
+                }
+                dir = dir.Parent;
+            }
+            if (fileInfo == null)
+            {
+                return null;
+            }
+            var t = new { sdk = new { version = "" } };
```
troydai: this is not used.

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,124 @@
...
+            try
+            {
+                var contents = File.ReadAllText(fileInfo.FullName);
+                var globalJson = JsonConvert.DeserializeObject<GlobalJsonStub>(contents);
+                return globalJson?.sdk?.version;
+            }
+            catch (JsonException)
+            {
+                // TODO log
```
troydai: it would be helpful at least the exception message is logged.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/InMemoryLogger.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Build.Framework;
+using System.Collections.Generic;
```
troydai: order

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/InMemoryLogger.cs
```diff
@@ -0,0 +1,48 @@
...
+    internal class InMemoryLogger : ILogger
+    {
+        private readonly Stack<Action> _onShutdown = new Stack<Action>();
+
+        internal IList<BuildWarningEventArgs> Warnings = new List<BuildWarningEventArgs>();
+        internal IList<BuildErrorEventArgs> Errors = new List<BuildErrorEventArgs>();
+
+        public string Parameters { get; set; }
+        public LoggerVerbosity Verbosity { get; set; }
+        public void Initialize(IEventSource eventSource)
```
troydai: add empty line between last property and the first method

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal class MsBuildContext
+    {
+        public string MsBuildExecutableFullPath { get; set; }
+        public string ExtensionsPath { get; set; }
+
+        public static MsBuildContext FromCurrentDotNetSdk()
+            => FromDotNetSdk(Path.GetDirectoryName(new Muxer().MuxerPath));
+
+        public static MsBuildContext FromDotNetSdk(string dotnetInstallationPath)
+            => FromDotNetSdk(/* projectDir */ null, dotnetInstallationPath);
```
troydai: use named parameter

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        internal static MsBuildContext FromDotNetSdk(string projectDir, string dotnetInstallationPath)
```
troydai: the first parameter is not used in any case.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    /// <summary>
+    /// Represents the msbuild context used to parse a project model
+    /// </summary>
+    internal class MsBuildContext
+    {
+        public string MsBuildExecutableFullPath { get; set; }
+        public string ExtensionsPath { get; set; }
```
troydai: can both properties be readonly?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        protected virtual void Initialize()
+        {
+            WithBuildTargets(new[] { "ResolveReferences" });
+            WithConfiguration("Debug");
+            WithProperty("_ResolveReferenceDependencies", "true");
+        }
+
+        protected virtual ProjectCollection CreateProjectCollection()
```
troydai: same line

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            Environment.SetEnvironmentVariable("MSBUILD_EXE_PATH", context.MsBuildExecutableFullPath);
+            return WithProperty("MSBuildExtensionsPath", context.ExtensionsPath);
+        }
+
+        public MsBuildProjectContextBuilder WithDesignTimeBuild()
+        {
+            // don't to expensive things
+            return WithProperty("DesignTimeBuild", "true")
+                     .WithProperty("_ResolveReferenceDependencies", "true")
```
troydai: my ocd tells me that the `W` should be aligned.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        public MsBuildProjectContextBuilder WithProperty(string property, string value)
+        {
+            _globalProperties[property] = value;
+            return this;
+        }
+
+        public MsBuildProjectContextBuilder WithBuildTargets(string[] targets)
+        {
```
troydai: check null?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+            ProjectCollection projectCollection)
+        {
+            using (var stream = fileInfo.CreateReadStream())
+            {
+                var xmlReader = XmlReader.Create(stream);
+
+                var xml = ProjectRootElement.Create(xmlReader, projectCollection, preserveFormatting: true);
+                xml.FullPath = fileInfo.PhysicalPath;
+
+                return new Project(xml, globalProps, /*toolsVersion:*/ null, projectCollection);
```
troydai: named parameter instead of comment among parameters

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.ProjectModel.Internal;
+using Microsoft.Extensions.FileProviders.Physical;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class MsBuildProjectContextBuilder
```
natemcmaster: What format does MSBuild expect TargetFramework to be? e.g. `.NETCoreApp,1.0` or `netcoreapp1.0`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
natemcmaster: Good point. I'll aggregate all errors.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
prafullbhosale: AFAIK `netcoreapp1.0`

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
natemcmaster: Nope, will clean this up.

makefile.shade
```diff
@@ -16,6 +16,7 @@ k-standard-goals
         if (Directory.Exists("src") && !IsLinux)
         {
           Directory.CreateDirectory(BUILD_DIR_LOCAL);
+          DotnetPack("src/Microsoft.Extensions.ProjectModel.Sources/project.json", BUILD_DIR_LOCAL, E("Configuration"), E("KOREBUILD_DOTNET_PACK_OPTIONS") + " --no-build");
```
natemcmaster: Because this makefile overrides the default build target.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (string.IsNullOrEmpty(configuration))
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            using (var stream = new FileStream(wrappedProject.ProjectFile.ProjectFilePath, FileMode.Open, FileAccess.Read))
+            using (var streamReader = new StreamReader(stream))
+            using (var jsonReader = new JsonTextReader(streamReader))
+            {
+                _rawProject = JObject.Load(jsonReader);
```
natemcmaster: I couldn't find API that exposes PJ as a grab bag of key/value pairs...but if it exists, happy to use it  instead.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
```diff
@@ -0,0 +1,101 @@
...
+                    return item.Value.Value<T>();
+                }
+            }
+            return default(T);
+        }
+
+        public IEnumerable<string> CompilationItems
+            => _compilerOptions.CompileInclude.ResolveFiles();
+        public IEnumerable<string> EmbededItems
+            => _compilerOptions.EmbedInclude.ResolveFiles();
```
natemcmaster: Yes, these are costly, but in theory can change over time. Callers should expect any evaluation of this yields the latest set of items and that each evaluation could yield different results.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContextFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.DotNet.ProjectModel;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class DotNetProjectContextFactory
+    {
+        public IProjectContext Create(string filePath, string configuration)
+        {
+            var project = new ProjectContextBuilder()
+                .AsDesignTime()
```
natemcmaster: IIUC this API ensures project.json => MSBuild references show up in list of dependencies.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/InMemoryLogger.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Build.Framework;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.ProjectModel.Internal
+{
+    internal class InMemoryLogger : ILogger
```
NTaylorMullen: Any reason it shouldn't be a private class in the class it's used in?

src/Microsoft.Extensions.ProjectModel.Sources/ProjectContextExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal static class ProjectContextExtensions
```
NTaylorMullen: These extensions seem wasteful. 

Reasons: 
- First is trying to avoid using ignorecase. Our dev guidelines state to always specify comparison.
- Second is overly specific to user secrets. Let the projects who care about user secrets add this API; no need to pollute the IntelliSense list with extra info when it typically will not be needed.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        private string _configuration;
+        private IFileInfo _fileInfo;
+        private string[] _buildTargets;
+        private Dictionary<string, string> _globalProperties = new Dictionary<string, string>();
+
+        public MsBuildProjectContextBuilder()
+        {
+            Initialize();
```
NTaylorMullen: lol, why do we need to specify a method that can be overridden (instead of inlining)? Makes me think the builder isn't sufficient as is.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        public MsBuildProjectContextBuilder WithProjectFile(string filePath)
+        {
+            if (string.IsNullOrEmpty(filePath))
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return WithProjectFile(new PhysicalFileInfo(new FileInfo(filePath)));
```
NTaylorMullen: break apart

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
NTaylorMullen: Seems like a lot here is protected virtual. So same question applies there.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public ProjectContext Unwrap() => _project;
```
natemcmaster: Good point. I'll removing so we are forced improve the IProjectContext abstraction instead of falling back. We can revisit later if there is a read need for this.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContextFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+                .WithProject(ProjectReader.GetProject(filePath))
+                .Build();
+
+            return new DotNetProjectContext(project, configuration, outputPath: null);
```
natemcmaster: Actually, I'm thinking of removing this factory altogether. I don't have a similar API for MSBuild, and I'm not sure the factory adds much value.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/ProjectContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.ProjectModel;
+
+namespace Microsoft.DotNet.ProjectModel
+{
+    internal static class ProjectContextExtensions
+    {
+        private const string DefaultConfiguration = "Debug";
+        public static IProjectContext AsAbstract(this ProjectContext context)
```
natemcmaster: There was like one use case when I started writing these abstractions, but I think that use case is gone now....so I'm removing this too.

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            if (latest == null)
+            {
+                throw CreateSdkNotInstalledException();
+            }
+
+            return new DotNetCoreSdk
+            {
+                BasePath = latest.path,
+                Version = latest.version.ToFullString()
```
natemcmaster: `NuGet.Versioning.SematicVersion` is an implementation detail of how we figure out which version is "latest". In theory, SDK versions may not be valid `SemanticVersion`s.

src/Microsoft.Extensions.ProjectModel.Sources/ProjectContextExtensions.cs
natemcmaster: Fair enough. I'm in favor of smaller APIs :)

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
natemcmaster: The intent is to make it possible to override builder defaults. Overriding methods is easier than unwinding work done in the base constructor.

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
NTaylorMullen: If they end up not being valid semantic versioning then doesn't our resolution logic fall apart? Reason being we order the versions based on how semantic versions are ordered.

natemcmaster:
:up: :date:

Made lots of updates per PR feedback.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
natemcmaster: Added `.WithTargetFramework`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
troydai: question is, do we need to. i know the `FindProperty` is defined on the interface, but all the information of a `project.json` is exposed through the project model.  the `project.json` shouldn't be treated as a simple dictionary.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContextFactory.cs
troydai: no. this ensures when fragment file is missing the project model will tolerant it. it doesn't add additional references, but allow them to be missing. see here: https://github.com/dotnet/cli/blob/1278e9d6bf35500e9672fe3f53125df1d08717d8/src/Microsoft.DotNet.ProjectModel/Resolution/MSBuildDependencyProvider.cs#L31

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
natemcmaster: > all the information of a project.json is exposed through the project model

Not user secrets id, or other custom properties...

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContextFactory.cs
natemcmaster: Ok. I removed this class anyways as the abstraction wasn't useful.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
troydai: Fair

troydai: Please still change the code style.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
```diff
@@ -0,0 +1,101 @@
...
+using Microsoft.DotNet.ProjectModel;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class DotNetProjectContext : IProjectContext
+    {
+        private readonly ProjectContext _project;
```
troydai: solved

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
troydai: fair.

src/Microsoft.Extensions.ProjectModel.Sources/Internal/DotNetCoreSdkResolver.cs
troydai: solved

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/InMemoryLogger.cs
troydai: solved

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildContext.cs
troydai: solved

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
troydai: and you fix that! thank you!

PR #181, opened 2016-10-05
Update ProjectModel to handle latest MSBuild updates and add MsBuildProjectFinder

natemcmaster:
Updating project model abstractions to test against the latest version of CLI/MSBuild. A few minor tweaks were required.

cc @NTaylorMullen @prafullbhosale can you review?


NTaylorMullen:
:shipit:

PR #184, opened 2016-10-06
Add api for multi-tfm projects

### LINKED ISSUE #183: Add support for multiple MSBuild target frameworks.
I'm not fully aware of the status of MSBuild supporting multiple target frameworks but it's something we'll need to make our ProjectModel aware of.

/cc @natemcmaster

natemcmaster:
Resolves #183

### DIFF
--- a/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
@@ -35,7 +35,19 @@ namespace Microsoft.Extensions.ProjectModel
         public string ProjectName => FindProperty("ProjectName") ?? _name;
         public string Configuration { get; }
 
-        public NuGetFramework TargetFramework => NuGetFramework.Parse(FindProperty("NuGetTargetMoniker"));
+        public NuGetFramework TargetFramework
+        {
+            get
+            {
+                var tfm = FindProperty("NuGetTargetMoniker") ?? FindProperty("TargetFramework");
+                if (tfm == null)
+                {
+                    return null;
+                }
+                return NuGetFramework.Parse(tfm);
+            }
+        }
+
         public bool IsClassLibrary => FindProperty("OutputType").Equals("Library", StringComparison.OrdinalIgnoreCase);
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
```diff
@@ -35,7 +35,19 @@ public string FindProperty(string propertyName)
         public string ProjectName => FindProperty("ProjectName") ?? _name;
         public string Configuration { get; }
 
-        public NuGetFramework TargetFramework => NuGetFramework.Parse(FindProperty("NuGetTargetMoniker"));
+        public NuGetFramework TargetFramework
```
NTaylorMullen: Would probably be easier to just swap this to `IEnumerable<NuGetFramework> TargetFrameworks`. That way consuming libraries can choose which framework to use.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -111,6 +111,22 @@ public virtual MsBuildProjectContext Build(bool ignoreBuildErrors = false)
             return new MsBuildProjectContext(name, _configuration, projectInstance);
         }
 
+        public IEnumerable<MsBuildProjectContext> BuildAllConfigurations()
```
NTaylorMullen: I'd change this to `MsBuildProjectContext Build(NuGetFramework framework)`.

No need to save the caller a for loop if not needed.

natemcmaster: There is a `.WithTargetFramework(NuGetFramework)` API already on the builder.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
NTaylorMullen: Ah I see. Still too used to old ProjectContext world from project.json 👍

PR #185, opened 2016-10-07
Add Dependency Resolution Apis to ProjectContext Abstractions

prafullbhosale:
Add Dependency Resolution Apis to ProjectContext Abstractions

@natemcmaster @NTaylorMullen

### DIFF
--- a/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
@@ -1,0 +1,98 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.IO;
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.Extensions.ProjectModel.Resolution;
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class DotNetDependencyProvider
+    {
+        private ProjectContext _context;
+        private List<DependencyDescription> _packageDependencies;
+        private List<ResolvedReference> _resolvedReferences;
+        private string _configuration;
+
+        public DotNetDependencyProvider(ProjectContext context, string configuration = Constants.DefaultConfiguration)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            _configuration = configuration;
+            _context = context;
+            DiscoverDependencies();
+        }
+
+        public IEnumerable<DependencyDescription> GetPackageDependencies()
+        {
+            return _packageDependencies;
+        }
+
+        public IEnumerable<ResolvedReference> GetResolvedReferences()
+        {
+            return _resolvedReferences;
+        }
+
+        private void DiscoverDependencies()
+        {
+            var exporter = _context.CreateExporter(_configuration);
+
+            if (exporter == null)
+            {
+                throw new InvalidOperationException($"Couldn't create a library exporter for configuration {_configuration}");
+            }
+
+            var framework = _context.TargetFramework;
+            if (framework == null)
+            {
+                throw new InvalidOperationException("Couldn't resolve dependencies when target framework is not specified.");
+            }
+
+            var exports = exporter.GetAllExports();
+            var nugetPackages = new Dictionary<string, DependencyDescription>(StringComparer.OrdinalIgnoreCase);
+            _resolvedReferences = new List<ResolvedReference>();
+
+            foreach (var export in exports)
+            {
+                var library = export.Library;
+
+                var description = new DependencyDescription(
+                    library.Identity.Name,
+                    library.Identity.Version.ToString(),
+                    export.Library.Path,
+                    framework.DotNetFrameworkName,
+                    library.Identity.Type.Value,
+                    library.Resolved);
+
+                foreach (var dependency in export.Library.Dependencies)
+                {
+                    var dep = new Dependency(dependency.Name, version: string.Empty);
+                    description.AddDependency(dep);
+                }
+
+                var itemSpec = $"{framework.DotNetFrameworkName}/{library.Identity.Name}/{library.Identity.Version.ToString()}";
+                nugetPackages[itemSpec] = description;
+
+                if (library.Resolved)
+                {
+                    foreach (var asset in export.CompilationAssemblies)
+                    {
+                        var resolvedRef = new ResolvedReference(
+                            name: Path.GetFileNameWithoutExtension(asset.FileName),
+                            resolvedPath: asset.ResolvedPath);
+                        _resolvedReferences.Add(resolvedRef);
+                    }
+                }
+            }
+
+            _packageDependencies = nugetPackages.Values.ToList();
+        }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using Microsoft.DotNet.ProjectModel;
+using Microsoft.Extensions.ProjectModel.Resolution;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using NuGet.Frameworks;
@@ -17,6 +18,10 @@ namespace Microsoft.Extensions.ProjectModel
         private readonly OutputPaths _paths;
         private readonly Lazy<JObject> _rawProject;
         private readonly CommonCompilerOptions _compilerOptions;
+        private readonly Lazy<DotNetDependencyProvider> _dependencyProvider;
+
+        private IEnumerable<DependencyDescription> _packageDependencies;
+        private IEnumerable<ResolvedReference> _compilationAssemblies;
 
         public DotNetProjectContext(ProjectContext projectContext, string configuration, string outputPath)
         {
@@ -49,6 +54,8 @@ namespace Microsoft.Extensions.ProjectModel
             // Workaround https://github.com/dotnet/cli/issues/3164
             IsClassLibrary = !(_compilerOptions.EmitEntryPoint
                     ?? projectContext.ProjectFile.GetCompilerOptions(null, configuration).EmitEntryPoint.GetValueOrDefault());
+
+            _dependencyProvider = new Lazy<DotNetDependencyProvider>(() => new DotNetDependencyProvider(_projectContext));
         }
 
         public bool IsClassLibrary { get; }
@@ -78,6 +85,32 @@ namespace Microsoft.Extensions.ProjectModel
 
         public IEnumerable<string> EmbededItems
             => _compilerOptions.EmbedInclude.ResolveFiles();
+
+        public IEnumerable<DependencyDescription> PackageDependencies
+        {
+            get
+            {
+                if (_packageDependencies == null)
+                {
+                    _packageDependencies = _dependencyProvider.Value.GetPackageDependencies();
+                }
+
+                return _packageDependencies;
+            }
+        }
+
+        public IEnumerable<ResolvedReference> CompilationAssemblies
+        {
+            get
+            {
+                if (_compilationAssemblies == null)
+                {
+                    _compilationAssemblies = _dependencyProvider.Value.GetResolvedReferences();
+                }
+
+                return _compilationAssemblies;
+            }
+        }
 
         /// <summary>
         /// Returns string values of top-level keys in the project.json file
--- a/src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using Microsoft.Extensions.ProjectModel.Resolution;
 using NuGet.Frameworks;
 
 namespace Microsoft.Extensions.ProjectModel
@@ -25,5 +26,7 @@ namespace Microsoft.Extensions.ProjectModel
         IEnumerable<string> CompilationItems { get; }
         IEnumerable<string> EmbededItems { get; }
         string FindProperty(string propertyName);
+        IEnumerable<DependencyDescription> PackageDependencies { get;}
+        IEnumerable<ResolvedReference> CompilationAssemblies { get; }
     }
 }
--- a/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
@@ -7,6 +7,7 @@ using System.IO;
 using Microsoft.Build.Execution;
 using NuGet.Frameworks;
 using System.Linq;
+using Microsoft.Extensions.ProjectModel.Resolution;
 
 namespace Microsoft.Extensions.ProjectModel
 {
@@ -16,23 +17,28 @@ namespace Microsoft.Extensions.ProjectModel
         private const string EmbedItemName = "EmbeddedResource";
         private const string FullPathMetadataName = "FullPath";
 
-        private readonly ProjectInstance _project;
-        private readonly string _name;
+        private readonly MsBuildProjectDependencyProvider _dependencyProvider;
+        private IEnumerable<DependencyDescription> _packageDependencies;
+        private IEnumerable<ResolvedReference> _compilationAssemblies;
+
+        protected ProjectInstance Project { get; }
+        protected string Name { get; }
 
         public MsBuildProjectContext(string name, string configuration, ProjectInstance project)
         {
-            _project = project;
+            Project = project;
 
             Configuration = configuration;
-            _name = name;
+            Name = name;
+            _dependencyProvider = new MsBuildProjectDependencyProvider(Project);
         }
 
         public string FindProperty(string propertyName)
         {
-            return _project.GetProperty(propertyName)?.EvaluatedValue;
+            return Project.GetProperty(propertyName)?.EvaluatedValue;
         }
 
-        public string ProjectName => FindProperty("ProjectName") ?? _name;
+        public string ProjectName => FindProperty("ProjectName") ?? Name;
         public string Configuration { get; }
 
         public NuGetFramework TargetFramework
@@ -69,14 +75,41 @@ namespace Microsoft.Extensions.ProjectModel
         }
         public string AssemblyFullPath => FindProperty("TargetPath");
         public string Platform => FindProperty("Platform");
-        public string ProjectFullPath => _project.FullPath;
+        public string ProjectFullPath => Project.FullPath;
         public string RootNamespace => FindProperty("RootNamespace") ?? ProjectName;
         public string TargetDirectory => FindProperty("TargetDir");
 
         public IEnumerable<string> CompilationItems
-            => _project.GetItems(CompileItemName).Select(i => i.GetMetadataValue(FullPathMetadataName));
+            => Project.GetItems(CompileItemName).Select(i => i.GetMetadataValue(FullPathMetadataName));
 
         public IEnumerable<string> EmbededItems
-             => _project.GetItems(EmbedItemName).Select(i => i.GetMetadataValue(FullPathMetadataName));
+             => Project.GetItems(EmbedItemName).Select(i => i.GetMetadataValue(FullPathMetadataName));
+
+        public IEnumerable<DependencyDescription> PackageDependencies
+        {
+            get
+            {
+                if (_packageDependencies == null)
+                {
+                    _packageDependencies = _dependencyProvider.GetPackageDependencies();
+                }
+
+                return _packageDependencies;
+            }
+        }
+
+        public IEnumerable<ResolvedReference> CompilationAssemblies
+        {
+            get
+            {
+                if (_compilationAssemblies == null)
+                {
+                    _compilationAssemblies = _dependencyProvider.GetResolvedReferences();
+                }
+
+                return _compilationAssemblies;
+            }
+        }
+
     }
 }
--- a/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
@@ -20,10 +20,10 @@ namespace Microsoft.Extensions.ProjectModel
 {
     internal class MsBuildProjectContextBuilder
     {
-        private string _configuration;
-        private IFileInfo _fileInfo;
-        private string[] _buildTargets;
-        private Dictionary<string, string> _globalProperties = new Dictionary<string, string>();
+        protected string _configuration;
+        protected IFileInfo _fileInfo;
+        protected string[] _buildTargets;
+        protected Dictionary<string, string> _globalProperties = new Dictionary<string, string>();
         private MsBuildContext _msbuildContext;
 
         public MsBuildProjectContextBuilder()
@@ -186,7 +186,7 @@ namespace Microsoft.Extensions.ProjectModel
 
         protected virtual void Initialize()
         {
-            WithBuildTargets(new[] { "ResolveReferences" });
+            WithBuildTargets(new[] { "ResolveReferences", "ResolvePackageDependenciesDesignTime" });
             WithProperty("_ResolveReferenceDependencies", "true");
         }
 
--- a/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
@@ -1,0 +1,126 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Execution;
+using Microsoft.Extensions.ProjectModel.Resolution;
+
+namespace Microsoft.Extensions.ProjectModel
+{
+    internal class MsBuildProjectDependencyProvider
+    {
+        private const string PackageDependencyItemType = "_DependenciesDesignTime";
+        private const string ResolvedReferenceItemType = "ReferencePath";
+
+        private readonly ProjectInstance _projectInstance;
+        public MsBuildProjectDependencyProvider(ProjectInstance projectInstance)
+        {
+            if (projectInstance == null)
+            {
+                throw new ArgumentNullException(nameof(projectInstance));
+            }
+            _projectInstance = projectInstance;
+        }
+
+        public IEnumerable<DependencyDescription> GetPackageDependencies()
+        {
+            var packageItems = _projectInstance.GetItems(PackageDependencyItemType);
+            var packageInfo = new Dictionary<string, DependencyDescription>(StringComparer.OrdinalIgnoreCase);
+            if (packageItems != null)
+            {
+                foreach (var packageItem in packageItems)
+                {
+                    var packageDependency = CreateDependencyDescriptionFromItem(packageItem);
+                    if (packageDependency != null)
+                    {
+                        packageInfo[packageItem.EvaluatedInclude] = packageDependency;
+                    }
+                }
+
+                // 2nd pass to populate dependencies;
+
+                PopulateDependencies(packageInfo, packageItems);
+            }
+
+            return packageInfo.Values;
+        }
+
+
+        public IEnumerable<ResolvedReference> GetResolvedReferences()
+        {
+            var refItems = _projectInstance.GetItems(ResolvedReferenceItemType);
+
+            var resolvedReferences = refItems
+                    ?.Select(refItem => CreateResolvedReferenceFromProjectItem(refItem))
+                    .Where(resolvedReference => resolvedReference != null);
+
+            return resolvedReferences;
+        }
+
+        private static ResolvedReference CreateResolvedReferenceFromProjectItem(ProjectItemInstance item)
+        {
+            var resolvedPath = item.EvaluatedInclude;
+
+            if (string.IsNullOrEmpty(resolvedPath))
+            {
+                return null;
+            }
+
+            var name = Path.GetFileNameWithoutExtension(resolvedPath);
+            return new ResolvedReference(name, resolvedPath);
+        }
+
+        private static DependencyDescription CreateDependencyDescriptionFromItem(ProjectItemInstance item)
+        {
+            // For type == Target, we do not get Name in the metadata. This is a special node where the dependencies are 
+            // the direct dependencies of the project.
+            var itemSpec = item.EvaluatedInclude;
+            var name = item.HasMetadata("Name") ? item.GetMetadataValue("Name") : itemSpec;
+
+            if (string.IsNullOrEmpty(name))
+            {
+                return null;
+            }
+
+            var version = item.GetMetadataValue("Version");
+            var path = item.GetMetadataValue("Path");
+            var type = item.GetMetadataValue("Type");
+            var resolved = item.GetMetadataValue("Resolved");
+
+            bool isResolved;
+            isResolved = bool.TryParse(resolved, out isResolved) ? isResolved : false;
+            var framework = itemSpec.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries).First();
+
+            return new DependencyDescription(name, version, path, framework, type, isResolved);
+        }
+
+        private static void PopulateDependencies(Dictionary<string, DependencyDescription> dependencies, ICollection<ProjectItemInstance> items)
+        {
+            foreach (var item in items)
+            {
+                var depSpecs = item.GetMetadataValue("Dependencies")
+                    ?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                DependencyDescription currentDescription = null;
+                if (depSpecs == null || !dependencies.TryGetValue(item.EvaluatedInclude, out currentDescription))
+                {
+                    return;
+                }
+
+                var prefix = item.EvaluatedInclude.Split('/').FirstOrDefault();
+                foreach (var depSpec in depSpecs)
+                {
+                    var spec =  $"{prefix}/{depSpec}";
+                    DependencyDescription dependency = null;
+                    if (dependencies.TryGetValue(spec, out dependency))
+                    {
+                        var dep = new Dependency(dependency.Name, dependency.Version);
+                        currentDescription.AddDependency(dep);
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/Dependency.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/Dependency.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class Dependency
+    {
+        public Dependency(string name, string version)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            Name = name;
+            Version = version;
+        }
+        public string Name { get; }
+        public string Version { get; }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class DependencyDescription
+    {
+        private readonly List<Dependency> _dependencies;
+
+        public DependencyDescription(string name, string version, string path, string framework, string type, bool isResolved)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (string.IsNullOrEmpty(framework))
+            {
+                throw new ArgumentNullException(nameof(framework));
+            }
+
+            Name = name;
+            Version = version;
+            TargetFramework = framework;
+            Resolved = isResolved;
+            Path = path;
+            DependencyType dt;
+            Type = Enum.TryParse(type, ignoreCase: true , result: out dt) ? dt : DependencyType.Unknown;
+
+            _dependencies = new List<Dependency>();
+        }
+
+        public string TargetFramework { get; }
+        public string Name { get; }
+        public string Path { get; }
+        public string Version { get; }
+        public DependencyType Type { get; }
+        public bool Resolved { get; }
+        public IEnumerable<Dependency> Dependencies => _dependencies;
+
+        public void AddDependency(Dependency dependency)
+        {
+            _dependencies.Add(dependency);
+        }
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyType.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyType.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public enum DependencyType
+    {
+        Target,
+        Package,
+        Assembly,
+        Project,
+        AnalyzerAssembly,
+        Unknown
+    }
+}
--- a/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/ResolvedReference.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/Resolution/ResolvedReference.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class ResolvedReference
+    {
+        public ResolvedReference(string name,string resolvedPath)
+        {
+            Name = name;
+            ResolvedPath = resolvedPath;
+        }
+        public string ResolvedPath { get; }
+        public string Name { get; }
+    }
+}
--- a/test/Microsoft.Extensions.ProjectModel.Tests/DotNet/DotNetDependencyProviderTests.cs
+++ b/test/Microsoft.Extensions.ProjectModel.Tests/DotNet/DotNetDependencyProviderTests.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.ProjectModel;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.ProjectModel.DotNet
+{
+    public class DotNetDependencyProviderTests
+    {
+        private const string projectJson = @"
+{
+  ""buildOptions"": {
+  },
+  ""dependencies"": {
+    ""Microsoft.AspNetCore.Mvc"": ""1.0.0-*"",
+  },
+  ""frameworks"": {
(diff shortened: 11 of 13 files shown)


natemcmaster:
Looks like this will need to be rebased

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
```diff
@@ -16,15 +17,18 @@ internal class MsBuildProjectContext : IProjectContext
         private const string EmbedItemName = "EmbeddedResource";
         private const string FullPathMetadataName = "FullPath";
 
-        private readonly ProjectInstance _project;
-        private readonly string _name;
+        protected readonly ProjectInstance _project;
```
natemcmaster: Can you rename to match our conventions for protected properties? 
` protected ProjectInstance Project { get; }`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            var framework = _context.TargetFramework;
+            if (exporter == null)
+            {
+                throw new ArgumentNullException(nameof(exporter));
+            }
+
+            if (framework == null)
+            {
+                throw new ArgumentNullException(nameof(framework));
```
natemcmaster: `InvalidOperationException("Couldn't resolve dependencies when target framework is not specified.")`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
```diff
@@ -49,6 +50,10 @@ public DotNetProjectContext(ProjectContext projectContext, string configuration,
             // Workaround https://github.com/dotnet/cli/issues/3164
             IsClassLibrary = !(_compilerOptions.EmitEntryPoint
                     ?? projectContext.ProjectFile.GetCompilerOptions(null, configuration).EmitEntryPoint.GetValueOrDefault());
+
+            var dependencyProvider = new DotNetDependencyProvider(_projectContext);
```
natemcmaster: This forces project context to have a TFM (because of the null check). Can we move this out of ctor and only instantiate when required?

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
```diff
@@ -24,5 +25,7 @@ public interface IProjectContext
         IEnumerable<string> CompilationItems { get; }
         IEnumerable<string> EmbededItems { get; }
         string FindProperty(string propertyName);
+        IEnumerable<DependencyDescription> Dependencies { get;}
+        IEnumerable<ResolvedReference> ResolvedReferences { get; }
```
natemcmaster: Can you clarify what's in this list? I'm guessing it's the sum of {ProjectReference, PackageReference, Reference (GAC or local) }

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,61 @@
...
+                throw new ArgumentNullException(nameof(framework));
+            }
+
+            Name = name;
+            Version = string.IsNullOrEmpty(version) ? DefaultVersion : version;
+            TargetFramework = framework;
+            Resolved = isResolved;
+            Path = path;
+            DependencyType dt;
+            Type = Enum.TryParse(type, out dt) ? dt : DependencyType.Unknown;
```
natemcmaster: Can you make this case insensitive? `Enum.TryParse(type, ignoreCase: true, out dt)`

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,61 @@
...
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if(string.IsNullOrEmpty(framework))
+            {
+                throw new ArgumentNullException(nameof(framework));
+            }
+
+            Name = name;
+            Version = string.IsNullOrEmpty(version) ? DefaultVersion : version;
```
natemcmaster: Can we just leave null? Or will bad things happen if we do?

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetProjectContext.cs
prafullbhosale: How about adding a builder method `.WithDependencyProvider` ? If there is no DependencyProvider specified, we don't expose any dependencies at all.

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
prafullbhosale: Yes, it is sum of all. But it points to the assemblies. Not to the packages/ projects.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            string itemSpec = string.Empty;
+            string nugetPackageSource = string.Empty;
+            string version = string.Empty;
+            string resolvedPath = string.Empty;
+            string name = string.Empty;
+
+            resolvedPath = item.EvaluatedInclude;
+
+            name = resolvedPath ?? Path.GetFileNameWithoutExtension(resolvedPath);
```
natemcmaster: This doesn't seem right. This will always throw if resolvedPath is null.

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
natemcmaster: Can we rename to `CompilationAssemblies`?

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
```diff
@@ -24,5 +25,7 @@ public interface IProjectContext
         IEnumerable<string> CompilationItems { get; }
         IEnumerable<string> EmbededItems { get; }
         string FindProperty(string propertyName);
+        IEnumerable<DependencyDescription> Dependencies { get;}
```
natemcmaster: How is this different from ResolvedReferences/CompilationAssemblies?

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
prafullbhosale: I suppose we could leave it null. 
Currently I don't have any scenarios that depend on versions.

natemcmaster: IMO we're better of keeping this projectmodel thing minimal, so probably leave out completely cause YAGNI,...but what do you think?

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
prafullbhosale: These are Package Dependencies. 
The dependency description also provides information like what are its dependencies. 
Scaffolding has a scenario where it looks for all the projects package dependencies which in turn have a dependency on a well known package. 

Another additional information that is missing from the ResolvedReferences collection is the path to the Nuget package folder for a given package, which might be useful if looking for non compilation assets

prafullbhosale: I think renaming the property here would make more sense. 
`DesignTimePackageDependencies` IMO would be more apt here.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            return new DependencyDescription(name, version, path, framework, type, isResolved);
+        }
+
+        private static void PopulateDependencies(Dictionary<string, DependencyDescription> dependencies, ICollection<ProjectItemInstance> items)
+        {
+            foreach (var item in items)
+            {
+                var depSpecs = item.GetMetadataValue("Dependencies")
+                    ?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                DependencyDescription current = null;
```
natemcmaster: `currentDescription`

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
natemcmaster: +1 `PackageDependencies`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                    var packageDependency = CreateDependencyDescriptionFromItem(packageItem);
+                    if (packageDependency != null)
+                    {
+                        packageInfo[packageItem.EvaluatedInclude] = packageDependency;
+                    }
+                }
+
+                // 2nd pass to populate dependencies;
+
+                PopulateDependencies(packageInfo, packageItems);
```
natemcmaster: This code is a little hard to follow. Can't this just be done in the first pass above, right after  CreateDependencyDescriptionFromItem? Or am I missing something.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            foreach (var export in exports)
+            {
+                var library = export.Library;
+                bool isResolved = bool.TryParse(library.Resolved.ToString(), out isResolved)
+                    ? isResolved
+                    : false;
+
+                var itemSpec = $"{framework.DotNetFrameworkName}/{library.Identity.Name}/{library.Identity.Version.ToString()}";
+                var description = new DependencyDescription(library.Identity.Name,
+                    export.Library.Path,
```
NTaylorMullen: These don't look to be in the right order. Ctor looks like: `public DependencyDescription(string name, string version, string path, string framework, string type, bool isResolved)`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                    itemSpec,
```
NTaylorMullen: So the DependencyDescription having the spec seems kind of odd. What's the purpose of this guy?

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                    library.Identity.Version.ToString(),
+                    library.Identity.Type.Value,
+                    isResolved);
+
+                foreach (var d in export.Library.Dependencies)
```
NTaylorMullen: d => dependency

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            var name = item.HasMetadata("Name") ? item.GetMetadataValue("Name") : itemSpec;
+
+            if (string.IsNullOrEmpty(name))
+            {
+                return null;
+            }
+
+            bool isResolved = false;
+            isResolved = bool.TryParse(resolved, out isResolved) ? isResolved : false;
+            var framework = itemSpec.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries).First();
```
natemcmaster: Since you just want the first part, this can be simplified to `framework = itemSpec.Substring(0, itemSpec.IndexOf('/'))`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    description.AddDependency(new Dependency(d.Name, d.VersionRange.MinVersion.ToString()));
```
NTaylorMullen: Split apart for debugging:

```C#
var dependency = new Dependency(dependency.Name, dependency.VersionRange.MinVersion.ToString())
description.AddDependency(dependency);
```

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
prafullbhosale: Well I'd keep it, just coz we don't need it now, doesn't mean we won't end up having a use case where we need to do something based on the version. 
Like with scaffolding, it not installs packages to user's project if they aren't already there. In future there could be a scenario, when it might need to upgrade a package if the version isn't correct.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            var version = item.GetMetadataValue("Version");
+            var path = item.GetMetadataValue("Path");
+            var type = item.GetMetadataValue("Type");
+            var resolved = item.GetMetadataValue("Resolved");
+            var itemSpec = item.EvaluatedInclude;
+
+            // For type == Target, we do not get Name in the metadata. This is a special node where the dependencies are 
+            // the direct dependencies of the project.
+
+            var name = item.HasMetadata("Name") ? item.GetMetadataValue("Name") : itemSpec;
```
natemcmaster: would just  `item.GetMetadataValue("Name")  ?? itemSpec` work?

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
NTaylorMullen: Is adding the MinVersion here sufficient? I know that VersionRange has an `IsMinInclusive` flag on it.

natemcmaster:
:watch: It's looking good so far. Requested some changes as above.

Also, can you add tests?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
prafullbhosale: AFAIK, `item.GetMetadataValue("Name")`  would return an empty string if the there is no metadata with key as `"Name"`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                nugetPackages[itemSpec] = description;
+
+                if (library.Resolved)
+                {
+                    foreach (var asset in export.CompilationAssemblies)
+                    {
+                        var resolvedRef = new ResolvedReference(asset.FileName, "1.0.0", asset.ResolvedPath, library.Identity.Type == LibraryType.Package);
```
NTaylorMullen: - 1.0.0 can't be right for the version here. Do we need the ResolvedReference category to have the version available? In project.json world the compilation assemblies never directly exposed that information on `LibraryAsset` which leads me to believe we don't need it now.

- Odd to have an isPackage identifier on a resolved reference. As far as a reference is concerned it's just an assembly, doesn't matter where it came from. Is that needed? Seems like it'd make more sense for a consumer to poke through the dependencies list to determine if a given assembly is a package (assuming they care).

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
natemcmaster: I'm okay leaving this. I can easily imagine uses for the API. But we should definitely not fill in a default version if it's not available. Let's leave null if the version is not given.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
natemcmaster: ok, then it's fine as is. MSBuild API is fickle. In other scenarios similar to this, it would return null instead of string.Empty

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
prafullbhosale: Every item here is represented using an 'itemspec' ( `item.EvaluatedInclude`)
This itemSpec is something like this (for a package)
`"TFMName/PackageName/PackageVersion"`

Lets say `Package A` has a dependency on `Package B`.
The way this is represented is 
The item representing `Package A` would have metadata with key as `Dependencies`
And the value for this metadata will be `"PackageName/PackageVersion"`. (as a `;` delimited string in case of multiple dependencies)

So to get more information regarding this, we need to lookup the item with itemspec 
`"TFMName/"` + whatever we got by parsing the Dependencies Metadata.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -113,7 +113,7 @@ public virtual MsBuildProjectContext Build(bool ignoreBuildErrors = false)
 
         protected virtual void Initialize()
         {
-            WithBuildTargets(new[] { "ResolveReferences" });
+            WithBuildTargets(new[] { "ResolveReferences", "ResolvePackageDependenciesDesignTime" });
```
NTaylorMullen: Curious, what's this guy do?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                throw new ArgumentNullException(nameof(projectInstance));
+            }
+            _projectInstance = projectInstance;
+        }
+
+        public IEnumerable<DependencyDescription> GetPackageDependencies()
+        {
+            var packageItems = _projectInstance.GetItems(PackageDependencyItemType);
+            var packageInfo = new Dictionary<string, DependencyDescription>();
```
NTaylorMullen: specify comparer

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
prafullbhosale: Although I haven't seen anything returned that has started with `/`, this would return an empty string. 

Also we do get one item returned here of type target, which won't have a `/` at all.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+
+            var isFromNugetPackage = "Package".Equals(nugetPackageSource, StringComparison.OrdinalIgnoreCase);
+            return new ResolvedReference(name, version, resolvedPath, isFromNugetPackage);
+        }
+
+        private static DependencyDescription CreateDependencyDescriptionFromItem(ProjectItemInstance item)
+        {
+            var version = item.GetMetadataValue("Version");
```
NTaylorMullen: Do all this resolution after your `if (string.IsNullOrEmpty(name)) return null` check

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            // the direct dependencies of the project.
+
+            var name = item.HasMetadata("Name") ? item.GetMetadataValue("Name") : itemSpec;
+
+            if (string.IsNullOrEmpty(name))
+            {
+                return null;
+            }
+
+            bool isResolved = false;
```
NTaylorMullen: `bool isResolved;`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                PopulateDependencies(packageInfo, packageItems);
+            }
+
+            return packageInfo.Values;
+        }
+
+
+        public IEnumerable<ResolvedReference> GetResolvedReferences()
+        {
+            var refItems = _projectInstance.GetItems(ResolvedReferenceItemType);
```
NTaylorMullen: Verifying: This includes all the references resolved from packages?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            var resolvedReferences = new List<ResolvedReference>();
+            if (refItems != null)
+            {
+                foreach (var refItem in refItems)
+                {
+                    resolvedReferences.Add(CreateResolvedReferenceFromProjectItem(refItem));
```
NTaylorMullen: Break into two lines to ease debugging

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            string nugetPackageSource = string.Empty;
+            string version = string.Empty;
+            string resolvedPath = string.Empty;
+            string name = string.Empty;
+
+            resolvedPath = item.EvaluatedInclude;
+
+            name = resolvedPath ?? Path.GetFileNameWithoutExtension(resolvedPath);
+
+            foreach (var m in item.Metadata)
```
NTaylorMullen: `var metadata in item.Metadata`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
natemcmaster: Thanks for clarifying. I'm trying to think of a way to simplify the code to make this more understandable, but can't think of any meaningful suggestions. :-/

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                if (ItemSpecKey.Equals(m.Name, StringComparison.OrdinalIgnoreCase))
```
NTaylorMullen: Pull these const bits into the method as variables.

Aka `const string ItemSpecKey = "OriginalItemSpec";`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
natemcmaster: Fair enough. Can you add tests for these different scenarios to ensure we get the right behavior?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                    ?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                DependencyDescription current = null;
+                if (depSpecs == null || !dependencies.TryGetValue(item.EvaluatedInclude, out current))
+                {
+                    return;
+                }
+
+                foreach (var depSpec in depSpecs)
+                {
+                    var spec = item.EvaluatedInclude.Split('/').FirstOrDefault()+ "/" + depSpec;
```
NTaylorMullen: - Move the prefix here outside of the foreach loop since the prefix doesn't change.
- Use IndexOf and substring instead of Split.
- `$"{prefix}/{depSpec}"` is your friend

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                    DependencyDescription d = null;
```
NTaylorMullen: don't be affraid to put the whole name. abbreviations are bad 😄

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/Dependency.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class Dependency
+    {
+        private const string DefaultVersion = "1.0.0";
```
NTaylorMullen: Why would we ever need a default version? If we're consuming a csproj it should have valid package versions specified.

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class DependencyDescription
+    {
+        private const string DefaultVersion = "1.0.0";
```
NTaylorMullen: Same question here as in Dependency

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,61 @@
...
+            _dependencies = new List<Dependency>();
+        }
+
+        public string TargetFramework { get; }
+        public string Name { get; }
+        public string Path { get; }
+        public string Version { get; }
+        public DependencyType Type { get; }
+        public bool Resolved { get; }
+        public IEnumerable<Dependency> Dependencies
```
NTaylorMullen: Why can't this be an `IEnumerable<DependencyDescription>`?

NTaylorMullen: Pranav points again `=>`

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
prafullbhosale: Not quite sure what the right thing here is. for now, we don't have any scenarios that lookup the version for the `Dependency`
So maybe it would be ok to keep it empty or remove it completely.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
prafullbhosale: You are right. 1.0.0 doesn't make sense.  Will keep it empty.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
prafullbhosale: packages, projects, GAC, assembly references...

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
prafullbhosale: Yeah, it doesn't really add any value to have isPackage. 
Will remove that.

prafullbhosale:
:up: :date: as per PR Feedback. 
Still working on tests.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContextBuilder.cs
```diff
@@ -19,11 +19,11 @@ namespace Microsoft.Extensions.ProjectModel
 {
     internal class MsBuildProjectContextBuilder
     {
-        private string _configuration;
-        private IFileInfo _fileInfo;
-        private string[] _buildTargets;
-        private Dictionary<string, string> _globalProperties = new Dictionary<string, string>();
-        private bool _explicitMsBuild;
+        protected string _configuration;
```
natemcmaster: Nit: always use properties with PascalCase names instead of protected fields 
`protected string Configuration { get; set; }`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        private static ResolvedReference CreateResolvedReferenceFromProjectItem(ProjectItemInstance item)
+        {
+            const string ItemSpecKey = "OriginalItemSpec";
+            const string NugetSourceTypeKey = "NugetSourceType";
+            const string VersionKey = "Version";
+
+
+            string itemSpec = string.Empty;
```
natemcmaster: `var`

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/ResolvedReference.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
natemcmaster: Nit: unused'using's.

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyType.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
natemcmaster: Nit: unused using's.

natemcmaster:
Good work on PR feedback. Looking a lot better now. I'll do more review when you have added tests.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            var exports = exporter.GetAllExports();
+            var nugetPackages = new Dictionary<string, DependencyDescription>(StringComparer.OrdinalIgnoreCase);
+            _resolvedReferences = new List<ResolvedReference>();
+
+            foreach (var export in exports)
+            {
+                var library = export.Library;
+                bool isResolved = library.Resolved;
```
NTaylorMullen: Just use library.Resolved elsewhere. No need for the variable

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+            resolvedPath = item.EvaluatedInclude;
+
+            name = string.IsNullOrEmpty(resolvedPath) ? "Unknown" : Path.GetFileNameWithoutExtension(resolvedPath);
+
+            foreach (var metadata in item.Metadata)
+            {
+                if (ItemSpecKey.Equals(metadata.Name, StringComparison.OrdinalIgnoreCase))
+                {
+                    itemSpec = metadata.EvaluatedValue;
+                    continue;
```
NTaylorMullen: Can just do elseif instead of continue.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                var itemSpec = $"{framework.DotNetFrameworkName}/{library.Identity.Name}/{library.Identity.Version.ToString()}";
```
NTaylorMullen: Move down to where it's used.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                nugetPackages[itemSpec] = description;
+
+                if (library.Resolved)
+                {
+                    foreach (var asset in export.CompilationAssemblies)
+                    {
+                        var resolvedRef = new ResolvedReference(
+                            asset.FileName,
+                            string.Empty,   // version
```
NTaylorMullen: named parameters 😄

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
NTaylorMullen: Do we need the version on ResolvedReference at all?

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
```diff
@@ -16,23 +17,29 @@ internal class MsBuildProjectContext : IProjectContext
...
 
         public MsBuildProjectContext(string name, string configuration, ProjectInstance project)
         {
-            _project = project;
+            Project = project;
 
             Configuration = configuration;
-            _name = name;
+            Name = name;
+            _dependencyProvider = new Lazy<MsBuildProjectDependencyProvider>(() =>
```
NTaylorMullen: Constructing `MsBuildProjectDependencyProvider` is incredibly cheap unless there's a bit in the implementation that I'm missing. Probably more expensive here to create a Lazy than the actual class 😉

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            return packageInfo.Values;
+        }
+
+
+        public IEnumerable<ResolvedReference> GetResolvedReferences()
+        {
+            var refItems = _projectInstance.GetItems(ResolvedReferenceItemType);
+            var resolvedReferences = new List<ResolvedReference>();
```
NTaylorMullen: This whole bit can be simplified to:
`var resolvedReferences = refItems?.Select(item => CreateResolvedReferenceFromProjectItem(item));`

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            return resolvedReferences;
+        }
+
+        private static ResolvedReference CreateResolvedReferenceFromProjectItem(ProjectItemInstance item)
+        {
+            const string ItemSpecKey = "OriginalItemSpec";
+            const string NugetSourceTypeKey = "NugetSourceType";
+            const string VersionKey = "Version";
+
```
NTaylorMullen: nit: extra line

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            string itemSpec = string.Empty;
+            string nugetPackageSource = string.Empty;
+            string version = string.Empty;
+            string resolvedPath = string.Empty;
+            string name = string.Empty;
```
NTaylorMullen: remove, you set this instantly on line 83

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectDependencyProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            resolvedPath = item.EvaluatedInclude;
+
+            name = string.IsNullOrEmpty(resolvedPath) ? "Unknown" : Path.GetFileNameWithoutExtension(resolvedPath);
```
prafullbhosale: Good point. I think, it would be more justifiable to just skip this item if we don't have a resolved path. 
Coz, if we don't have resolved path, the only thing we would have is Version.

prafullbhosale:
:up: :date: 
Added tests for Dependency Provider

test/Microsoft.Extensions.ProjectModel.Tests/DotNet/DotNetDependencyProviderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        <add key=""nugetbuild"" value=""https://www.myget.org/F/nugetbuild/api/v3/index.json"" />
+    </packageSources>
+</configuration>";
+
+        private const string projectJson = @"
+{
+  ""buildOptions"": {
+  },
+  ""dependencies"": {
+    ""Microsoft.AspNetCore.Mvc"": ""1.0.0-*"",
```
natemcmaster: Can you add this as a PackageReference instead? PJ is dead

natemcmaster: Nevermind, this is intentionally a PJ test....

prafullbhosale: This is test for Abstraction using the Project.json. 
The MsBuild related tests are in the MsBuildDependencyProviderTests.

test/Microsoft.Extensions.ProjectModel.Tests/DotNet/DotNetDependencyProviderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            _output = output;
+        }
+
+        [Fact]
+        public void BuildProjectDependencies()
+        {
+            using (var fileProvider = new TemporaryFileProvider())
+            {
+                Directory.CreateDirectory(Path.Combine(fileProvider.Root, "demo"));
+                fileProvider.Add("NuGet.config", NugetConfigTxt);
```
natemcmaster: Is this required for this test, or can we just restore stable packages?

test/Microsoft.Extensions.ProjectModel.Tests/MsBuild/MsBuildProjectDependencyProviderTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                fileProvider.Add($"Root{Path.DirectorySeparatorChar}Two.cs", "public class Abc2 {}");
+                fileProvider.Add($"Root{Path.DirectorySeparatorChar}Excluded.cs", "public class Abc {}");
+
+                // Add Class Library project
+                fileProvider.Add($"Library1{Path.DirectorySeparatorChar}library.csproj", LibraryProjectTxt);
+                fileProvider.Add($"Library1{Path.DirectorySeparatorChar}Three.cs", "public class Abc3 {}");
+
+                var testContext = _fixture.GetMsBuildContext();
+
+                var muxer = Path.Combine(testContext.ExtensionsPath, "../..", "dotnet.exe");
```
natemcmaster: On linux, the muxer doesn't end in .exe....but we can resolve that later when we remove the skip for this test.

test/Microsoft.Extensions.ProjectModel.Tests/MsBuild/MsBuildProjectDependencyProviderTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                var result = Command
+                    .Create(muxer, new[] { "restore3", Path.Combine(fileProvider.Root, "Library1","Library1.csproj") })
```
natemcmaster: Is it possible to do restore on a solution file instead of on each project? (I wouldn't be surprised if that doesn't work yet.)

test/Microsoft.Extensions.ProjectModel.Tests/MsBuild/MsBuildProjectDependencyProviderTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                // This reference doesn't resolve so should not be available here.
+                Assert.False(compilationAssemblies.Any(assembly => assembly.Name.Equals("xyz", StringComparison.OrdinalIgnoreCase)));
+
+                var packageDependencies = context.PackageDependencies;
+                var mvcPackage = packageDependencies
+                    .Where(p => p.Name.Equals("Microsoft.AspNetCore.Mvc", StringComparison.OrdinalIgnoreCase))
+                    .FirstOrDefault();
+                Assert.NotNull(mvcPackage);
+
+                Assert.True(mvcPackage.Dependencies.Any(dependency => dependency.Name.Equals("Microsoft.Extensions.DependencyInjection", StringComparison.OrdinalIgnoreCase)));
```
natemcmaster: Has the same effect, but you could also use
 `Assert.Contains(mvcPackage.Dependencies, d => d.Name.Equals("Microsoft.Extensions.DependencyInjection"))`

natemcmaster:
Looks good. :shipit:

test/Microsoft.Extensions.ProjectModel.Tests/MsBuild/MsBuildProjectDependencyProviderTests.cs
prafullbhosale: I haven't really tried restoring on a solution recently.

prafullbhosale:
I have rebased and squashed the commits. But don't have write access to the repo.

NTaylorMullen:
What about: https://github.com/aspnet/DotNetTools/pull/185#discussion_r82480276

natemcmaster:
> But don't have write access to the repo.

Ask Eilon for access. 

Also, my :shipit: is just for my concerns. I think some of @NTaylorMullen's requests are still unresolved.

NTaylorMullen:
> But don't have write access to the repo.
> Ask Eilon for access.

@Eilon 😄

prafullbhosale:
Got the access already!

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
NTaylorMullen: Ya lets nuke it. Reason being, the project.json world never supported it (`LibraryAsset` never gave you access to the version).

prafullbhosale: Done.

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        public IEnumerable<ResolvedReference> GetResolvedReferences()
+        {
+            return _resolvedReferences;
+        }
+
+        private void DiscoverDependencies()
+        {
+            var exporter = _context.CreateExporter(_configuration);
+
+            var framework = _context.TargetFramework;
```
NTaylorMullen: move down to where you null check. Confusing here

src/Microsoft.Extensions.ProjectModel.Sources/DotNet/DotNetDependencyProvider.cs
NTaylorMullen: Can just use library.Resolved. Don't need the variable.

src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
```diff
@@ -16,23 +17,28 @@ internal class MsBuildProjectContext : IProjectContext
         private const string EmbedItemName = "EmbeddedResource";
         private const string FullPathMetadataName = "FullPath";
 
-        private readonly ProjectInstance _project;
-        private readonly string _name;
+        private MsBuildProjectDependencyProvider _dependencyProvider;
```
NTaylorMullen: readonly

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.ProjectModel.Resolution
+{
+    public class DependencyDescription
+    {
+        private List<Dependency> _dependencies;
```
NTaylorMullen: readonly

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public DependencyDescription(string name, string version, string path, string framework, string type, bool isResolved)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if(string.IsNullOrEmpty(framework))
```
NTaylorMullen: nit: `if (`

src/Microsoft.Extensions.ProjectModel.Sources/Resolution/DependencyDescription.cs
```diff
@@ -0,0 +1,54 @@
...
+        public string Name { get; }
+        public string Path { get; }
+        public string Version { get; }
+        public DependencyType Type { get; }
+        public bool Resolved { get; }
+        public IEnumerable<Dependency> Dependencies => _dependencies;
+
+        public void AddDependency(Dependency dependency)
+        {
+            if (dependency != null)
```
NTaylorMullen: No need to protect here. If a user calls AddDependency with null they're shooting themselves in the foot.

PR #187, opened 2016-10-10
[small] Add PackageLockFile to IProjectContext

natemcmaster:
dotnet-watch will need this.

### DIFF
--- a/src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Extensions.ProjectModel
         string Config { get; }
         string DepsJson { get; }
         string RuntimeConfigJson { get; }
+        string PackageLockFile { get; }
         string PackagesDirectory { get; }
         string TargetDirectory { get; }
         string AssemblyFullPath { get; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.ProjectModel.Sources/MsBuild/MsBuildProjectContext.cs
```diff
@@ -54,7 +54,7 @@ public NuGetFramework TargetFramework
         public string Config => AssemblyFullPath + ".config";
         public string DepsJson => Path.Combine(TargetDirectory, Path.GetFileNameWithoutExtension(AssemblyFullPath) + ".deps.json");
         public string RuntimeConfigJson => Path.Combine(TargetDirectory, Path.GetFileNameWithoutExtension(AssemblyFullPath) + ".runtimeconfig.json");
-
+        public string PackageLockFile => Path.Combine(Path.GetDirectoryName(ProjectFullPath), "obj", "project.assets.json");
```
prafullbhosale: Would this path ever change based on any properties in the project?

natemcmaster: Good point. Dug a little deeper. Looks like it could be in a different folder.

cref https://github.com/NuGet/NuGet.Client/blob/971f47f885db1afcb7845411df38b1212e4151b0/src/NuGet.Core/NuGet.ProjectModel/RestoreOutputType.cs#L15-L18
https://github.com/NuGet/NuGet.Client/blob/1c1d537b379936a76dcc2b010fdfdecb54657070/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L315

src/Microsoft.Extensions.ProjectModel.Sources/IProjectContext.cs
```diff
@@ -18,6 +18,7 @@ public interface IProjectContext
         string Config { get; }
         string DepsJson { get; }
         string RuntimeConfigJson { get; }
+        string PackageLockFile { get; }
```
NTaylorMullen: If we wanted we could make this property smarter in the sense that it could validate that the file exists, if not returns `null`. Super minimal and I'm not even that convinced it's worth it. Food for thought unless you felt different.

natemcmaster: >  I'm not even that convinced it's worth it. 

Me neither. It's really just mean to find the file, not necessarily to read/parse it.

PR #189, opened 2016-10-11
Remove use of MSBuild API directly in dotnet-user-secrets

natemcmaster:
Use indirect project evaluation. Avoids ~~possible~~ numerous problems with using MSBuild API directly.

This will invoke msbuild indirectly to find the UserSecretsId property.

cc @prafullbhosale @NTaylorMullen @jeffkl


src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    internal class MsBuildProjectFinder
+    {
+        private readonly string _directory;
+
+        public MsBuildProjectFinder(string directory)
+        {
+            if (string.IsNullOrEmpty(directory))
+            {
+                throw new ArgumentException("Value cannot be null or empty", nameof(directory));
```
NTaylorMullen: Resource

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            _directory = directory;
+        }
+
+        public string FindMsBuildProject(string project = null)
```
NTaylorMullen: No need for default parameter.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (!File.Exists(projectPath))
+            {
+                throw FileDoesNotExist(projectPath);
+            }
+
+            return projectPath;
+        }
+
+        private Exception FileDoesNotExist(string filePath)
```
NTaylorMullen: No longer a need for these methods, just inline em.

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -147,27 +143,80 @@ internal int RunInternal(params string[] args)
...
-                var project = new MsBuildProjectContextBuilder()
-                    .UseMsBuild(_msBuildContext)
-                    .AsDesignTimeBuild()
-                    .WithBuildTargets(Array.Empty<string>())
-                    .WithProjectFile(projectFile)
-                    .WithTargetFramework("") // TFM doesn't matter
-                    .Build();
-
-                return project.GetUserSecretsId();
+                var commandResult = Command.CreateDotNet("msbuild",
```
NTaylorMullen: Assuming we get buy in from the cli folks this will eventually be called for us, and then this EXE will be called after correct?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -147,27 +143,80 @@ internal int RunInternal(params string[] args)
...
+                    throw new GracefulException(Resources.FormatError_ProjectMissingId(projectFile));
+                }
+
+                return id;
+            }
+            finally
+            {
+                if (File.Exists(outputFile))
+                {
+                    File.Delete(outputFile);
```
NTaylorMullen: Should try catch this deletion, don't care if it fails.

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -147,27 +143,80 @@ internal int RunInternal(params string[] args)
...
+
+            if (foundFile != null)
             {
-                throw new GracefulException(Resources.FormatError_ProjectFailedToLoad(projectFile), ex);
+                return foundFile;
             }
+
+            // back up plan - if the targets file was not found any of the expected locations,
+            // create the file using the embedded version
+            var outputPath = Path.Combine(AppContext.BaseDirectory, targetFileName);
```
NTaylorMullen: This is an interesting support path. What's the purpose here?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
natemcmaster: In theory, that could work, but I'm not planning on anything like that unless we can get a design from CLI folks that offers clear benefits.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -0,0 +1,58 @@
...
+            var projectPath = project ?? _directory;
+
+            if (!Path.IsPathRooted(projectPath))
+            {
+                projectPath = Path.Combine(_directory, projectPath);
+            }
+
+            if (Directory.Exists(projectPath))
+            {
+                var projects = Directory.EnumerateFileSystemEntries(projectPath, "*.*proj", SearchOption.TopDirectoryOnly).ToList();
```
prafullbhosale: Will this not find .xproj files as well?

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
natemcmaster: Currently used in testing. This can probably be removed later, but it's necessary as there isn't (yet) a good way to get the .targets file into the test project once we delete the src folder.

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -0,0 +1,129 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal
+{
+    public class ProjectIdResolver : IDisposable
+    {
+        private const string TargetsFileName = "FindUserSecretsProperty.targets";
+        private readonly ILogger _logger;
+        private readonly string _workingDirectory;
+        private Stack<Action> _disposables = new Stack<Action>();
```
prafullbhosale: What does this indicate? 
Perhaps a better name would make it clear?

natemcmaster: It's a collection of items to be disposed. See the .Dispose() method.

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
natemcmaster: oh, yeah, it will. I'll add a filter.

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
NTaylorMullen: readonly

NTaylorMullen: `Stack<string>` of files to delete. Could even just use list. Could also rename to something like  `_createdFiles`

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -0,0 +1,129 @@
...
+                    targetFile,
+                    "/nologo",
+                    "/t:_FindUserSecretsProperty",
+                    $"/p:Project={projectFile}",
+                    $"/p:OutputFile={outputFile}",
+                    $"/p:Configuration={configuration}"
+                })
+                .CaptureStdErr()
+                .CaptureStdOut()
+                .OnErrorLine(l => commandOutput.Add(l))
```
NTaylorMullen: Can't you just forward error and output to a string buffer? It'd save you from string.Joining below

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -0,0 +1,129 @@
...
+                throw new GracefulException(Resources.FormatError_ProjectFailedToLoad(projectFile));
+            }
+
+            var id = File.ReadAllText(outputFile);
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new GracefulException(Resources.FormatError_ProjectMissingId(projectFile));
+            }
+
+            return id.Trim();
```
NTaylorMullen: If you're going to Trim do it above with a `?.Trim()` so you also catch pure whitespace

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -0,0 +1,129 @@
...
+                resource.CopyTo(stream);
+            }
+
+            // cleanup
+            _disposables.Push(() => TryDelete(outputPath));
+
+            return outputPath;
+        }
+
+        private static void TryDelete(string file)
```
NTaylorMullen: When you go to a string for the disposables just inline this method in the dispose

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -0,0 +1,129 @@
...
+                .Select(dir => Path.Combine(dir, TargetsFileName))
+                .Where(File.Exists)
+                .FirstOrDefault();
+
+            if (foundFile != null)
+            {
+                return foundFile;
+            }
+
+            // back up plan - if the targets file was not found any of the expected locations,
```
NTaylorMullen: Since this is used today for testing/non prod scenarios make that clear in the comment for future devs

src/Microsoft.Extensions.SecretManager.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -33,15 +33,18 @@ public string FindMsBuildProject(string project)
 
             if (Directory.Exists(projectPath))
             {
-                var projects = Directory.EnumerateFileSystemEntries(projectPath, "*.*proj", SearchOption.TopDirectoryOnly).ToList();
+                var projects = Directory.EnumerateFileSystemEntries(projectPath, "*.*proj", SearchOption.TopDirectoryOnly)
+                    .Where(f => !".xproj".Equals(Path.GetExtension(f), StringComparison.OrdinalIgnoreCase))
```
prafullbhosale: Instead of filtering out .xproj, why not just allow the extensions that you know about?

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
natemcmaster: UX choice. I don't want to show the MSBuild output unless there is a problem with it.

NTaylorMullen: kk

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
natemcmaster: What's wrong with current implementation?

NTaylorMullen: Why have an action that creates scopes when all you're doing is deleting created files?

natemcmaster: I prefer the flexiblity thie current structure lends to the code. Also, it's CLI tool and the perf gain is so minimal it's not noticeable, may not even be measurable compared to how slow invoking a new MSBuild process will be. 

But not worth spending more cycles on debating small stuff like this. We have bigger fish to fry. Will update.

jeffkl:
@natemcmaster it _might_ be cool if the `FindUserSecretsProperty.targets` file was an embedded resource that gets written to `%TEMP%` at runtime.  Just a thought.

FYI @rainersigwald and @AndyGerlicher.  I wonder if this sort of thing should ship with MSBuild?

PR #191, opened 2016-10-12
Handle compilerOptions in project.json

prafullbhosale:
Currently if the project.json doesn't have the 
```
"buildOptions": {
  "compile": "*.cs"
}
```
the CompilationItems will throw Exception.
Same for Embedded items as well. 

cc @natemcmaster @NTaylorMullen

### DIFF
--- a/src/Microsoft.Extensions.ProjectModel.DotNet.Sources/DotNetProjectContext.cs
+++ b/src/Microsoft.Extensions.ProjectModel.DotNet.Sources/DotNetProjectContext.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Extensions.ProjectModel
             _compilerOptions = _projectContext.ProjectFile.GetCompilerOptions(TargetFramework, Configuration);
 
             // Workaround https://github.com/dotnet/cli/issues/3164
-            IsClassLibrary = !(_compilerOptions.EmitEntryPoint
+            IsClassLibrary = !(_compilerOptions?.EmitEntryPoint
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.ProjectModel.DotNet.Sources/DotNetProjectContext.cs
```diff
@@ -83,10 +83,14 @@ public DotNetProjectContext(ProjectContext projectContext, string configuration,
         public string Platform => _compilerOptions.Platform;
```
prafullbhosale: Not sure what the default here should be.

natemcmaster: `_compilerOptions?.Platform` is fine

src/Microsoft.Extensions.ProjectModel.DotNet.Sources/DotNetProjectContext.cs
```diff
@@ -83,10 +83,14 @@ public DotNetProjectContext(ProjectContext projectContext, string configuration,
...
 
         public IEnumerable<string> CompilationItems
-            => _compilerOptions.CompileInclude.ResolveFiles();
+            => (_compilerOptions == null || _compilerOptions.CompileInclude == null)
```
natemcmaster: Optional nit: possible to shorten to `_compilerOptions?.CompileInclude == null`

NTaylorMullen:
:shipit:

PR #194, opened 2016-10-12
Simplify the end to end testing to not rely on tool installation via restore

### LINKED ISSUE #193: dotnet-watch tests not guaranteed to test the version of dotnet-watch in the source
dotnet-watch functional tests run a package restore to put the tool into the test projects. In some cases, this means the tool being tested is not actually the version of the tool that was built during the test run. When running on a local box, this will often test the latest version on our nightly feeds, not the version I'm actually working on.

natemcmaster:
Resolves #193

cc @moozzyk


test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/Scenario/ProjectToolScenario.cs
```diff
@@ -26,57 +32,34 @@ public ProjectToolScenario()
...
-            var tempNugetConfigFile = Path.Combine(WorkFolder, NugetConfigFileName);
-
-            var nugetConfig = XDocument.Load(tempNugetConfigFile);
-            var packageSource = nugetConfig.Element("configuration").Element("packageSources");
-            packageSource.Add(new XElement("add", new XAttribute("key", feedName), new XAttribute("value", feed)));
-            using (var stream = File.OpenWrite(tempNugetConfigFile))
-            {
-                nugetConfig.Save(stream);
+                var dest = file.Replace(srcFolder, destinationFolder);
+                if ("P.json".Equals(Path.GetFileName(file), StringComparison.OrdinalIgnoreCase))
```
moozzyk: This is weird - was the file renamed to `P.json` to prevent restoring dependencies?

moozzyk: Does it add too much to the build? I wonder it would not be better just to restore for these projects instead of introducing P.json which no one will be familiar with.

natemcmaster: Hmm, fair point. Tried it: it doesn't really seem to cause problems when named project.json so I'll undo that rename.

moozzyk:
:ship::it:

PR #195, opened 2016-10-12
Update to netcoreapp1.1

natemcmaster:
Finish up work partially done in fd6df3c74dd77a1e167ef1361816323946497719

cc @pranavkm


natemcmaster:
Will also need to update https://github.com/aspnet/Coherence-Signed/blob/dev/tools/NugetPackageVerifier.json

natemcmaster:
Nevermind. We're holding off on this change.

PR #196, opened 2016-10-13
Refactor dotnet-watch to isolate project.json dependency

natemcmaster:
Changes:

- General code cleanup
- Remove interfaces when there isn't more than one implementation.
- Fixup code that mixes of thread blocking and task awaiting code.
- Decouple file watching for understanding of project model.

cc @moozzyk


src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
```diff
@@ -1,176 +1,80 @@
...
-                {
-                    return;
-                }
+                        _logger.LogInformation("Waiting for a file to change before restarting dotnet...");
 
-                _logger.LogError($"Error(s) reading project file '{projectFile}': ");
-                _logger.LogError(errors);
-                _logger.LogInformation("Fix the error to continue.");
+                        // Now wait for a file to change before restarting process
+                        await Task.WhenAll(fileSetTask).ConfigureAwait(false);
```
moozzyk: There is just one task - can't you just do `await fileSetTask.ConfigureAwait(false)` ?

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
```diff
@@ -1,176 +1,80 @@
...
-                    int finishedTaskIndex = Task.WaitAny(tasksToWait, cancellationToken);
-
-                    // Regardless of the outcome, make sure everything is cancelled
+                    // Regardless of the which task finished first, make sure everything is cancelled
                     // and wait for dotnet to exit. We don't want orphan processes
                     currentRunCancellationSource.Cancel();
-                    Task.WaitAll(tasksToWait);
 
-                    cancellationToken.ThrowIfCancellationRequested();
+                    await Task.WhenAll(processTask, fileSetTask).ConfigureAwait(false);
```
moozzyk: If the user kills the process won't we block here until a file is changed as well?

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
moozzyk: ...and you probably would have to use `.Result` on the next line.

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
```diff
@@ -1,176 +1,80 @@
...
-                throw new ArgumentNullException(nameof(cancellationToken));
-            }
+            Ensure.NotNull(processSpec, nameof(processSpec));
+            Ensure.NotNull(cancellationToken, nameof(cancellationToken));
 
-            var dotnetArgumentsAsString = ArgumentEscaper.EscapeAndConcatenateArgArrayForProcessStart(dotnetArguments);
-
-            var workingDir = Path.GetDirectoryName(projectFile);
+            var cancelledTaskSource = new TaskCompletionSource<object>();
+            cancellationToken.Register(state => ((TaskCompletionSource<object>)state).TrySetResult(null), cancelledTaskSource);
```
moozzyk: Is there any specific reason you use the overload with `state` here?

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // because .NET Core doesn't have the Process.Disposed event and .Exited isn't always raised
+            while (!process.HasExited)
+            {
+                await Task.WhenAny(Task.Delay(500), tcs.Task).ConfigureAwait(false);
+            }
+
+            var exitCode = process.ExitCode;
+            process.Dispose();
+            process = null; // set null so cancellation token callback doesn't matter
```
moozzyk: Is this safe (threading)?

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,94 @@
...
+            EventHandler handler = (sender, args) =>
+            {
+                tcs.TrySetResult(null);
+            };
+
+            process.Exited += handler;
+            
+            cancellationToken.Register(() =>
+            {
+                process?.KillTree();
```
moozzyk: maybe check `process.HasExited`? `process?` seems reacey.

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,94 @@
...
+            return exitCode;
+        }
+
+        private Process CreateProcess(ProcessSpec processSpec)
+        {
+            var arguments = ArgumentEscaper.EscapeAndConcatenateArgArrayForProcessStart(processSpec.Arguments);
+
+            _logger.LogInformation("Running {execName} with the following arguments: {args}", processSpec.ShortDisplayName(), arguments);
+
+            var startInfo = new ProcessStartInfo()
```
moozzyk: nit: remove `()`

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                FileName = processSpec.Executable,
+                Arguments = arguments,
+                UseShellExecute = false,
+                WorkingDirectory = processSpec.WorkingDirectory
+            };
+            var process = new Process()
```
moozzyk: ditto

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
moozzyk: actually process.HasExited would not work. Maybe assign to a local variable first and then do `.?` ?

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -14,10 +14,7 @@ internal class CommandLineOptions
         public IList<string> RemainingArguments { get; private set; }
         public static CommandLineOptions Parse(string[] args, TextWriter consoleOutput)
         {
-            if (args == null)
-            {
-                throw new ArgumentNullException(nameof(args));
-            }
+            Ensure.NotNull(args, nameof(args));
```
natemcmaster: [EF does this.](https://github.com/aspnet/EntityFramework/blob/dev/src/Shared/Check.cs). I like the terseness, and I'mnNot too worried about the stack trace issue as this is "product code" in the sense that dotnet-watch isn't a class library people will use. But if you'd still rather we don't use this pattern here, I can revert.

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
```diff
@@ -1,176 +1,80 @@
...
-            if (dotnetArguments == null)
-            {
-                throw new ArgumentNullException(nameof(dotnetArguments));
-            }
-            if (cancellationToken == null)
-            {
-                throw new ArgumentNullException(nameof(cancellationToken));
-            }
+            Ensure.NotNull(processSpec, nameof(processSpec));
+            Ensure.NotNull(cancellationToken, nameof(cancellationToken));
```
natemcmaster: I guess this means the previous version didn't work either...will update.

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
natemcmaster: Embarassing.... yeah, that can be cleaned up.

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
natemcmaster: Yeah, there is a race problem. :watch: on trying to figure this out.

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
moozzyk: This correct. I actually thought the compiler returns a warning in this case.

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
natemcmaster: Yeah, that's the design. If process dies or is killed, we block on a file change to relaunch.

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            var wasKilled = false;
+            var hasExited = false; // used to avoid race condition with process.Dispose()
+
+            cancellationToken.Register(() =>
+            {
+                if (!hasExited && !process.HasExited) 
+                {
+                    wasKilled = true;
+                    process.KillTree();
```
moozzyk: can this throw?

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,109 @@
...
+            Ensure.NotNull(logger, nameof(logger));
+
+            _logger = logger;
+        }
+
+        public async Task<int> RunAsync(ProcessSpec processSpec, CancellationToken cancellationToken)
+        {
+            Ensure.NotNull(processSpec, nameof(processSpec));
+
+            var killLock = new object();
```
moozzyk: not used

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                tcs.TrySetResult(null);
+            };
+
+            var process = CreateProcess(processSpec);
+
+            process.Exited += handler;
+
+            var wasKilled = false;
+            var hasExited = false; // used to avoid race condition with process.Dispose()
```
moozzyk: I don't see how it helps avoid a race condition - cancellation token fires while you are waiting in line 59. You go to line 47 and then the other thread resumes and the RunAsync runs to the end. The you are back on this thread and try to kill the process tree but the process has already been disposed. Also note that `wasKilled` flag in this scenario is also incorrect. 
I am not sure how helpful/interesting the `wasKilled` flag is but I would remove it - it makes this code even more complicated and I am not sure if it even is possible to get it always right. 
For the handler itself - I think it is fine to try to prevent killing the process if it has already exited but you have a couple non-atomic operations here some of them out of your control so you may just want to check if the process has not exited and then try to kill it if it hasn't and catch/swallow all the exceptions since there is not much you can do.

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                if (!hasExited && !process.HasExited) 
+                {
+                    wasKilled = true;
+                    process.KillTree();
+                }
+                tcs.TrySetResult(null);
+            });
+
+            process.Start();
```
moozzyk: What happens if this throws? Does it completely kill dotnet watch?

src/Microsoft.DotNet.Watcher.Tools/Internal/Implementation/ProjectReader.cs
```diff
@@ -36,11 +34,11 @@ public bool TryReadProject(string projectFile, out IProject project, out string
         }
 
         // Same as TryGetProject but it doesn't throw
```
natemcmaster: Yeah, i'm not sure what it means either. Looks like the comment has survived since days of dnx-watch. Probably can be cleaned out.

image

src/Microsoft.DotNet.Watcher.Tools/Internal/Implementation/ProjectReader.cs
```diff
@@ -2,16 +2,14 @@
...
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using Microsoft.DotNet.ProjectModel;
 
 namespace Microsoft.DotNet.Watcher.Internal
 {
-    public class ProjectProvider : IProjectProvider
+    public class ProjectReader
```
natemcmaster: +1 ProjectReaderUtils

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
natemcmaster: Yeah, it can throw, but I don't think most of the conditions apply. https://msdn.microsoft.com/en-us/library/0w4h05yb(v=vs.110).aspx#Anchor_1. But if we start supporting arbitrary executable.

natemcmaster:
:up: :date: to catch possible exceptions in process.KillTree. Rebased on dev.

Issue #142, opened 2016-07-21
Cannot specify a log level from the CLI


NotMyself:
I am attempting to write an interactive tutorial and want to use dotnet-watch. I would like to suppress the info messages printed out by the watch command so that only messages from my tutorial are displayed.

I have tried adding a logging.json file, but it doesn't seem to be respected. The log level in the Tools project appears to be [hard coded to info](https://github.com/aspnet/dotnet-watch/blob/dev/src/Microsoft.DotNet.Watcher.Tools/CommandOutputProvider.cs#L27). How can I override it?

victorhurdugaci:
Agree, it would be a nice feature. It's a bit tricky to implement because we need a way to pass arguments to `dotnet watch`. Today, it just passes everything to `dotnet`. We'll definitely investigate this

NotMyself:
It might be as simple as looking for a logging.json so it can be configured externally.

muratg:
Assigning to @victorhurdugaci for invesigation in 1.1.0.

victorhurdugaci:
Hey @glennc, what do you think about this one?

glennc:
hmmm. What are examples of the output that you want to suppress? Is it all output from watch?

I am thinking a switch passed to watch to stop it from logging anything. `-[q]uite`

victorhurdugaci:
Talked to @glennc offline. We'll make this configurable through and environment variable, just like we do for enabling the polling watcher.

NotMyself:
Here is an example from the [DotNetCoreKoans](https://github.com/NotMyself/DotNetCoreKoans) project. The project is a teaching tool where the student is instructed to open the project in a text editor and then run `dotnet watch run` in their console. The student then changes code in a set of unit tests learning about the C# language. The output of the application guides them  through each tutorial.

When running the application without the watcher, I have complete control over the output. With the watcher, it clutters up the output making it hard to follow. 

At a minimum it would be nice to have the watcher respect a config file or a flag to specify silent output.

![screen shot 2016-08-30 at 3 16 11 pm](https://cloud.githubusercontent.com/assets/73120/18109185/bae0c3da-6ec4-11e6-928b-bfc8e3af1a10.png)

muratg:
Would we be still displaying errors in "quiet" mode, and just be silent otherwise? Or completely silent?

victorhurdugaci:
[...]

PR #197, opened 2016-10-13
dotnet-watch: add --verbose and --quiet command line flags

natemcmaster:
Command line options are more intuitive than an environment variable. This adds two command line options as an alternate solution for https://github.com/aspnet/DotNetTools/issues/142

cc @glennc @moozzyk @muratg


src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -44,8 +51,15 @@ public static CommandLineOptions Parse(string[] args, TextWriter consoleOutput)
...
 
+            if (optQuiet.HasValue() && optVerbose.HasValue())
+            {
+                consoleOutput.WriteLine("Found both --quiet and --verbose options. The --verbose flag will be ignored.".Bold().Black());
+            }
+
             return new CommandLineOptions
             {
+                IsQuiet = optQuiet.HasValue(),
+                IsVerbose = optVerbose.HasValue() && !optQuiet.HasValue(), // --quiet takes precedence
```
muratg: Should "verbose" override "quite"?  @glennc thoughts?

natemcmaster:
To clarify, this is an alternate **solution**, not an alternate configuration option. This PR would actually remove the environment variable for configuring log level.

src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -88,6 +72,18 @@ public static int Main(string[] args)
                 return 2;
             }
 
+            var logLevel = options.IsQuiet
+                ? LogLevel.Warning
+                : options.IsVerbose
+                    ? LogLevel.Debug
+                    : LogLevel.Information;
+
+            var commandProvider = new CommandOutputProvider()
```
moozzyk: nit: remove `()`

src/Microsoft.DotNet.Watcher.Tools/Program.cs
```diff
@@ -29,22 +29,6 @@ public Program(TextWriter consoleOutput, CancellationToken cancellationToken)
 
             _cancellationToken = cancellationToken;
             _out = consoleOutput;
-
-            _loggerFactory = new LoggerFactory();
-
-            var logVar = Environment.GetEnvironmentVariable("DOTNET_WATCH_LOG_LEVEL");
```
moozzyk: I don't know why we used environment variable but removing it could break things (I am mostly concerned about some kind of tooling)

moozzyk: In that case I think it is fine to remove this.

moozzyk:
:ship::it:

src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
glennc: Should it be an error case instead? You can't be noisy and quite at the same time so show an error saying what the problem is.

natemcmaster: Yeah, that's a better approach. Will update to that.

natemcmaster:
Question from an after thought: dotnet supports a "global verbose" flag, which we currently ignore. Should `dotnet --verbose watch` be considered equivalent to `dotnet watch --verbose`?

natemcmaster:
Furthermore, if `dotnet watch --verbose` is set, should watch set the global verbose flag?

i.e. `dotnet watch --verbose run` would launch the equivalent of `dotnet --verbose run`

PR #199, opened 2016-10-15
dotnet-user-secrets: add support for json output and piping json input

### LINKED ISSUE #168: Make tool more friendly in usage from other tooling.
_From @phenning on August 24, 2016 20:9_

For example, in order to display a grid, you would need to parse out the info: output from the results of dotnet user-secrets list. Ideally, this should just be a --json or similar switch which would just output the secrets as json.

Similarly, it would also be good to be able to pass in json via the set command, bonus if nested json could be passed in here and the tool would take care of flattening. 

_Copied from original issue: aspnet/UserSecrets#95_

natemcmaster:
Resolves #168.

This adds support for `dotnet user-secrets list --json` which will spit out JSON.
It also adds support for piping input to the "set" command, e.g. `cat secrets.json | dotnet user-secrets set`

cc @phenning @muratg @glennc @moozzyk

### DIFF
--- a/src/Microsoft.Extensions.SecretManager.Tools/Internal/SetCommand.cs
+++ b/src/Microsoft.Extensions.SecretManager.Tools/Internal/SetCommand.cs
@@ -1,12 +1,15 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Diagnostics;
+using System.IO;
+using System.Text;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Internal
 {
-    internal class SetCommand : ICommand
+    public class SetCommand : ICommand
     {
         private readonly string _keyName;
         private readonly string _keyValue;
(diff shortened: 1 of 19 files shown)


natemcmaster:
Can someone review this?

src/Microsoft.Extensions.SecretManager.Tools/CommandContext.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal
+{
+    internal class CommandContext
```
muratg: I'm OK keeping this internal, though the usual pattern I believe is to keep the classes public, and indicate that they're internal via the namespace.

src/Microsoft.Extensions.SecretManager.Tools/Internal/SetCommand.cs
```diff
@@ -11,41 +14,90 @@ internal class SetCommand : ICommand
         private readonly string _keyName;
         private readonly string _keyValue;
 
-        public static void Configure(CommandLineApplication command, CommandLineOptions options)
+        public static void Configure(ExtendedHelpCommandLineApplication command, CommandLineOptions options)
         {
             command.Description = "Sets the user secret to the specified value";
+            command.AdditionalHelpInfo = @"
+Additional Info:
+  This command will also handled piped input. Piped input is expected to be a valid JSON format.
```
muratg: nit: tpyo: handled -> handle

src/Microsoft.Extensions.SecretManager.Tools/README.md
```diff
@@ -0,0 +1,22 @@
+dotnet-user-secrets
```
muratg: 👍

PR #200, opened 2016-10-17
Merge feature/msbuild to dev

natemcmaster:
Now that we've branched the preview3 release, we should be able to do our msbuild work on dev branch, right?

cc @pranavkm @NTaylorMullen @bricelam @prafullbhosale  @muratg


NTaylorMullen:
If we merged it, it'd definitely be temporary given our "long" term plan for tools.

natemcmaster:
Premature. Our CI isn't ready for this. We need to take preview3 build with MSBuild support for this work, but the rest of our runtime bits aren't quite ready to build on preview3.

PR #201, opened 2016-10-17
Upgrade to MSBuild 15.1.319-preview5 and fix broken tests.

natemcmaster:
cc @pranavkm  should unblock CI.
cc @prafullbhosale some of the tests you wrote needed changing.

### DIFF
--- a/test/Microsoft.Extensions.ProjectModel.Tests/project.json
+++ b/test/Microsoft.Extensions.ProjectModel.Tests/project.json
@@ -7,10 +7,11 @@
     }
   },
   "dependencies": {
-    "NuGet.Frameworks": "3.5.0-rc1-final",
+    "NuGet.ProjectModel": "3.6.0-*",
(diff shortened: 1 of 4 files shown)


test/Microsoft.Extensions.ProjectModel.Tests/project.json
```diff
@@ -7,10 +7,11 @@
     }
   },
   "dependencies": {
-    "NuGet.Frameworks": "3.5.0-rc1-final",
+    "NuGet.ProjectModel": "3.6.0-*",
```
natemcmaster: @prafullbhosale @NTaylorMullen just ran into the problem that @davkean predicted. Project evaluation is through MSBuild APIs is going to be really hard to do right because as soon as the SDK changes their references, our project evaluation fails.  We should try to get off using MSBuild API directly as soon as possible.

NTaylorMullen:
:shipit:

PR #213, opened 2016-11-02
Bundle deps.json files in tools packages

natemcmaster:
Attempt to future proof the packages against https://github.com/NuGet/Home/issues/3673


NTaylorMullen:
:shipit:

natemcmaster:
:watch: actually, from talking with @rrelyea we released we need the deps.json from dotnet-publish, not dotnet-build. Will update.

natemcmaster:
:up: 📅

PR #215, opened 2016-11-03
Lift tools versions to 1.1.0

natemcmaster:
Tools that require .NET Core 1.1 should version as 1.1.

cc @muratg @Eilon @shanselman


muratg:
:shipit: 

Could you also share versioning plan (the table you had) with  all the tool owners so that they also make the necessary versioning updates?

Issue #205, opened 2016-10-19
dotnet-watch: show example usages in help message


cesarbs:
Current message:

`Usage: dotnet watch [options] [[--] <arg>...]`

Should be

`Usage: dotnet watch COMMAND [options] [[--] <arg>...]`

Command being `run`, `test` i.e. a `dotnet` command.

natemcmaster:
*Technically, the current message is accurate, but I'm guessing you (and others) won't realize that `[[--] <arg>...]` means "put run/test/etc here". Probably could be helped by adding an example to the help message.

cesarbs:
I don't really get that from `[[--] <arg>...]`. That part of the usage message is saying "if you want to pass args to the command being watched, use `--` followed by the command's args.

natemcmaster:
The command goes after the (optional) `--` separator. Examples of correct usage:
`dotnet watch run arg1 arg2`
`dotnet watch -- run arg1 arg2`
`dotnet watch -- test -class A.B.Test`.

FWIW, it's matches what the CLI does:

![image](https://cloud.githubusercontent.com/assets/2696087/19569795/de6d32c0-96ab-11e6-9101-3f8379fd4f32.png)

But I think this is a good ask: the default help message isn't super obvious

cesarbs:
That's not what's documented here: https://github.com/aspnet/DotNetTools/tree/dev/src/Microsoft.DotNet.Watcher.Tools

natemcmaster:
I don't see the difference. Let's talk in person.

muratg:
Looks like this was done. Good to close?

natemcmaster:
No, it's not done yet.

PR #218, opened 2016-11-03
dotnet-watch: add extended help to clarify usage

natemcmaster:
Resolves https://github.com/aspnet/DotNetTools/issues/205

cc @cesarbs

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
@@ -29,9 +29,17 @@ namespace Microsoft.DotNet.Watcher
                 AllowArgumentSeparator = true,
                 ExtendedHelpText = @"
 Remarks:
-  The argument separator '--' can be used to ensure separation of the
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Watcher.Tools/CommandLineOptions.cs
```diff
@@ -26,7 +26,17 @@ public static CommandLineOptions Parse(string[] args, TextWriter stdout, TextWri
...
+                AllowArgumentSeparator = true,
+                ExtendedHelpText = @"
+Remarks:
+  The argument separator '--' can be used to ensure separation of the
+  arguments for 'dotnet-watch' and those passed into the subprocess.
+  For example: dotnet watch --quiet -- test -class TestClass1
+
+Examples:
+  dotnet watch run
+  dotnet watch test
```
cesarbs: I think this is a little confusing in that in the first example it says `-- test`, but here it's just `test`. How could you make it clearer that both work the same way?

natemcmaster:
:up: :date:

Issue #224, opened 2016-11-07
dotnet-watch restarts app but does not recompile when files are deleted


natemcmaster:
**Repro**:

1. Create a files that uses globs to import *.cs files.
1. dotnet watch run
1. Delete a file matched by the glob

**Expected**
Dotnet-watch restarts app with deleted file no longer included in the build.

**Actual**
dotnet-watch restarts the app, but the app was not recompiled.

**Cause:**
https://github.com/Microsoft/msbuild/issues/701

natemcmaster:
Fixed in dotnet-cli 1.0.0-preview4-004233

Issue #212, opened 2016-11-02
dotnet-watch for MSBuild does not restart process when new file is added


natemcmaster:
**Repro:**
watch an MSBuild project that uses globs.
Add a new file that would match one of the globs.

**Expected**
dotnet-watch should restart process.

**Actual**
new file is ignored

**Remarks**
We should revisit the testing as the tests supposedly test this scenario and are passing, but this does not actually work.

natemcmaster:
cc @muratg I thought this was a simple fix, but when I tried to implement it, everything blew up. I'm not sure I'll get this in for 1.1.0 some I'm opening this to track progress.

muratg:
The impact is not huge, because once you edit another existing file (to use the new file contents) build is started. Backlogging for now.

natemcmaster:
I propose we close. We've had no reports of this being an issue.

natemcmaster:
User reported this in https://github.com/dotnet/cli/issues/9886. Re-opening to reconsider.

PR #225, opened 2016-11-07
Fix false positive test passes

natemcmaster:
The end to end tests for dotnet-watch were passing when they should not have been due to race conditions in the test.

Bugs: 
https://github.com/aspnet/DotNetTools/issues/212
https://github.com/aspnet/DotNetTools/issues/224

cc @moozzyk


natemcmaster:
Ping. @moozzyk or @cesarbs could you review?

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
```diff
@@ -0,0 +1,123 @@
...
+        private void StartProcessingOutput(StreamReader streamReader)
+        {
+            _source = _source ?? new BufferBlock<string>();
+            Interlocked.Increment(ref _reading);
+            Task.Run(() =>
+            {
+                var buffer = new char[1];
+                var sb = new StringBuilder();
+                var lastChar = '\0';
+                while (streamReader.Read(buffer, 0, 1) > 0)
```
moozzyk: why not just ReadLine?

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        private void StartProcessingOutput(StreamReader streamReader)
+        {
+            _source = _source ?? new BufferBlock<string>();
```
moozzyk: for some reason I feel uneasy with `BufferBlock` - feels like something from a different world...

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/TestProjects/GlobbingApp/Program.cs
```diff
@@ -3,37 +3,19 @@
...
-        private static void Block()
-        {
-            while (true)
-            {
-                ConsoleWrite("Blocked...");
-                Thread.Sleep(1000);
-            }
+            Console.WriteLine("Started");
+            Console.WriteLine("PID = " + Process.GetCurrentProcess().Id);
+            Console.WriteLine("Defined types = " + typeof(Program).GetTypeInfo().Assembly.DefinedTypes.Count());
```
moozzyk: See it now in a test..

cesarbs:
Everything looks ok to me, but this is a fairly large change. Can you elaborate on what exactly was wrong with the tests?

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
```diff
@@ -2,132 +2,63 @@
...
-                    File.WriteAllText(fileToChange, programCs);
-
-                    wait.Wait(_defaultTimeout,
-                        expectedToChange: true,
-                        errorMessage: $"Process did not restart because {scenario.StartedFile} was not changed");
-                }
-            }
+            await _app.StartWatcher().OrTimeout();
+            var pid = await _app.GetProcessId().OrTimeout();
+            await _app.HasExited().OrTimeout(); // process should exit after run
```
moozzyk: Remind me - the process is not running after this line until the file is changed when watcher will restart the process and let it exit - is this correct?

natemcmaster:
The tests were designed to use files on disk for IPC, but this kind of IPC had race conditions.

Example:
1. Xunit starts dotnet-watch which will run test-app
2. test-app writes to a file 'started.txt'
3. xunit waits for changes to 'started.txt' to change. When it does, it attempts to change a file and see if dotnet-watch will restart test-app.

The problem with this approach was step 3 used a polling file watcher. Depending on when the poll happens, 'started.txt' could represent the first time test-app runs, or the second time test-app has run.
One solution could have been to add additional waits and checks to ensure the change to 'started.txt' represented what we wanted, but this got messy.

These changes simplify the whole test design by instead just reading stdout from the dotnet-watch process in order to determine what is happening inside this process.

cesarbs:
Cool. 🚢 when Travis is green.

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/NoDepsAppTests.cs
natemcmaster: Yes, that's right. If the process dotnet-watch is monitoring exits on its own, dotnet-watch will still hang around and wait for file changes before it restarts the process.

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
natemcmaster: It seemed like a good way to do async producer/consumer. Are there potential issues with this API?

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/TestProjects/GlobbingApp/Program.cs
natemcmaster: yeah, it's meant to help tests inspect if the app recompiled with(out) the file changes.

Issue #227, opened 2016-11-12
Make log format consistent across tools


cesarbs:
E.g. when I try to use `sql-cache` or `user-secrets` with an invalid argument, the error format is different between the two:

**sql-cache**

```

$ dotnet sql-cache --foo
Specify --help for a list of available options and commands.
crit: Microsoft.Extensions.Caching.SqlConfig.Tools.Program[0]
      An error occurred. Unrecognized option '--foo'
```

**user-secrets**

```
$ dotnet user-secrets --foo
Specify --help for a list of available options and commands.
critical: Command failed : Unrecognized option '--foo'
```

`watch` errors have a different format too. For example:

```
[DotNetWatcher] fail: dotnet exit code: 134
```

natemcmaster:
Also, we should make usage of colors consistent.

natemcmaster:
Fixed in #237

PR #232, opened 2016-11-17
dotnet-watch: remove dependency on Microsoft.DotNet.Cli.Utils

natemcmaster:
dotnet-watch doesn't really need Microsoft.DotNet.Cli.Utils. This PR trims the dependency because dotnet-watch uses only a very small portion of the API.

This will also allow us to (potentially) create dotnet-watch for net451 (Microsoft.DotNet.Cli.Utils requires net46).

cc @prafullbhosale @cesarbs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ project.lock.json
 testWorkDir/
 *.nuget.props
 *.nuget.targets
+.idea/
--- a/src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
@@ -5,6 +5,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.Watcher.Internal;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Tools.Internal;
 
 namespace Microsoft.DotNet.Watcher
 {
@@ -44,6 +45,10 @@ namespace Microsoft.DotNet.Watcher
                 {
                     var fileSetTask = fileSetWatcher.GetChangedFileAsync(combinedCancellationSource.Token);
                     var processTask = _processRunner.RunAsync(processSpec, combinedCancellationSource.Token);
+
+                    _logger.LogInformation("Running {execName} with the following arguments: {args}",
+                        processSpec.ShortDisplayName(),
+                        ArgumentEscaper.EscapeAndConcatenate(processSpec.Arguments));
 
                     var finishedTask = await Task.WhenAny(processTask, fileSetTask, cancelledTaskSource.Task);
 
--- a/src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
@@ -7,11 +7,11 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.DotNet.Cli.Utils;
-using Microsoft.DotNet.Watcher.Tools;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Tools.Internal;
 
 namespace Microsoft.DotNet.Watcher.Internal
 {
@@ -23,6 +23,7 @@ namespace Microsoft.DotNet.Watcher.Internal
         private readonly string _projectFile;
         private readonly string _watchTargetsDir;
         private readonly OutputSink _outputSink;
+        private readonly ProcessRunner _processRunner;
 
         public MsBuildFileSetFactory(ILogger logger, string projectFile)
             : this(logger, projectFile, new OutputSink())
@@ -40,6 +41,7 @@ namespace Microsoft.DotNet.Watcher.Internal
             _projectFile = projectFile;
             _watchTargetsDir = FindWatchTargetsDir();
             _outputSink = outputSink;
+            _processRunner = new ProcessRunner(logger);
         }
 
         internal List<string> BuildFlags { get; } = new List<string>
@@ -71,20 +73,21 @@ namespace Microsoft.DotNet.Watcher.Internal
                     var capture = _outputSink.StartCapture();
                     // TODO adding files doesn't currently work. Need to provide a way to detect new files
                     // find files
-                    var exitCode = Command.CreateDotNet("msbuild",
-                        new[]
+                    var processSpec = new ProcessSpec
+                    {
+                        Executable = DotNetMuxer.MuxerPathOrDefault(),
+                        WorkingDirectory = projectDir,
+                        Arguments = new[]
                         {
-                            _projectFile,
+                            "msbuild",
+                             _projectFile,
                             $"/p:_DotNetWatchTargetsLocation={_watchTargetsDir}", // add our dotnet-watch targets
-                            $"/p:_DotNetWatchListFile={watchList}",
-                        }.Concat(BuildFlags))
-                        .CaptureStdErr()
-                        .CaptureStdOut()
-                        .OnErrorLine(l => capture.WriteErrorLine(l))
-                        .OnOutputLine(l => capture.WriteOutputLine(l))
-                        .WorkingDirectory(projectDir)
-                        .Execute()
-                        .ExitCode;
+                            $"/p:_DotNetWatchListFile={watchList}"
+                        }.Concat(BuildFlags),
+                        OutputCapture = capture
+                    };
+
+                    var exitCode = await _processRunner.RunAsync(processSpec, cancellationToken);
 
                     if (exitCode == 0)
                     {
@@ -104,9 +107,18 @@ namespace Microsoft.DotNet.Watcher.Internal
                         return fileset;
                     }
 
-                    _logger.LogError($"Error(s) finding watch items project file '{Path.GetFileName(_projectFile)}': ");
-                    _logger.LogError(capture.GetAllLines("[MSBUILD] : "));
-                    _logger.LogInformation("Fix the error to continue.");
+                    var sb = new StringBuilder()
+                        .Append("Error(s) finding watch items project file '")
+                        .Append(Path.GetFileName(_projectFile))
+                        .AppendLine("' :");
+
+                    foreach (var line in capture.Lines)
+                    {
+                        sb.Append("  [MSBUILD] :").AppendLine(line);
+                    }
+
+                    _logger.LogError(sb.ToString());
+                    _logger.LogInformation("Fix the error to continue or press Ctrl+C to exit.");
 
                     var fileSet = new FileSet(new[] { _projectFile });
 
--- a/src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildProjectFinder.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildProjectFinder.cs
@@ -4,7 +4,6 @@
 using System;
 using System.IO;
 using System.Linq;
-using Microsoft.DotNet.Cli.Utils;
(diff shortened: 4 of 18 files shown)


.gitignore
```diff
@@ -31,4 +31,5 @@ project.lock.json
 .vscode/
 testWorkDir/
 *.nuget.props
-*.nuget.targets
\ No newline at end of file
+*.nuget.targets
+.idea/
```
cesarbs: What's this?

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
```diff
@@ -45,6 +46,10 @@ public DotNetWatcher(ILogger logger)
                     var fileSetTask = fileSetWatcher.GetChangedFileAsync(combinedCancellationSource.Token);
                     var processTask = _processRunner.RunAsync(processSpec, combinedCancellationSource.Token);
 
+                    _logger.LogInformation("Running {execName} with the following arguments: {args}",
+                        processSpec.ShortDisplayName(),
+                        ArgumentEscaper.EscapeAndConcatenate(processSpec.Arguments));
```
cesarbs: What's the difference between doing this an interpolating args in the string itself (`$""`)?

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -104,9 +107,16 @@ internal MsBuildFileSetFactory(ILogger logger, string projectFile, OutputSink ou
                         return fileset;
                     }
 
-                    _logger.LogError($"Error(s) finding watch items project file '{Path.GetFileName(_projectFile)}': ");
-                    _logger.LogError(capture.GetAllLines("[MSBUILD] : "));
-                    _logger.LogInformation("Fix the error to continue.");
+                    var sb = new StringBuilder()
+                        .Append("Error(s) finding watch items project file '")
+                        .Append(Path.GetFileName(_projectFile))
```
cesarbs: You could just interpolate this in the string, no?

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -71,20 +73,21 @@ internal MsBuildFileSetFactory(ILogger logger, string projectFile, OutputSink ou
...
-                        .CaptureStdOut()
-                        .OnErrorLine(l => capture.WriteErrorLine(l))
-                        .OnOutputLine(l => capture.WriteOutputLine(l))
-                        .WorkingDirectory(projectDir)
-                        .Execute()
-                        .ExitCode;
+                            $"/p:_DotNetWatchListFile={watchList}"
+                        }.Concat(BuildFlags),
+                        OutputCapture = capture
+                    };
```
cesarbs: Is this copied from `Command.CreateDotNet()`? I'm concerned the original code contains workarounds and whatnot that we could be missing here.

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -104,9 +107,16 @@ internal MsBuildFileSetFactory(ILogger logger, string projectFile, OutputSink ou
...
-                    _logger.LogError(capture.GetAllLines("[MSBUILD] : "));
-                    _logger.LogInformation("Fix the error to continue.");
+                    var sb = new StringBuilder()
+                        .Append("Error(s) finding watch items project file '")
+                        .Append(Path.GetFileName(_projectFile))
+                        .AppendLine("' :");
+                    foreach (var line in capture.Lines)
+                    {
+                        sb.Append("  [MSBUILD] :").AppendLine(line);
+                    }
```
cesarbs: nit: add blank lines before and after the `foreach` block.

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -104,9 +107,16 @@ internal MsBuildFileSetFactory(ILogger logger, string projectFile, OutputSink ou
...
+                    _logger.LogError(sb.ToString());
+                    _logger.LogInformation("Fix the error to continue or press CTRL+C to exit.");
```
cesarbs: Current notation is to use `Ctrl`.

src/Microsoft.DotNet.Watcher.Tools/Internal/ProcessRunner.cs
```diff
@@ -35,30 +36,43 @@ public ProcessRunner(ILogger logger)
                 cancellationToken.Register(() => processState.TryKill());
 
                 process.Start();
-                _logger.LogInformation("{execName} process id: {pid}", processSpec.ShortDisplayName(), process.Id);
 
-                await processState.Task;
+                if (processSpec.IsOutputCaptured)
+                {
+
```
cesarbs: Remove blank line.

src/Shared/ArgumentEscaper.cs
```diff
@@ -0,0 +1,128 @@
...
+        private static IEnumerable<string> EscapeArgArray(IEnumerable<string> args)
+        {
+            var escapedArgs = new List<string>();
+
+            foreach (var arg in args)
+            {
+                escapedArgs.Add(EscapeSingleArg(arg));
+            }
+
+            return escapedArgs;
```
cesarbs: `=> args.Select(EscapeSingleArg)`

+`ToArray()`/`ToList()` if you want to avoid running the state machine more than once when you iterate over the result.

src/Shared/ArgumentEscaper.cs
```diff
@@ -0,0 +1,128 @@
...
+                while (i < arg.Length && arg[i] == '\\')
+                {
+                    backslashCount++;
+                    i++;
+                }
+
+                // Escape any backslashes at the end of the arg
+                // when the argument is also quoted.
+                // This ensures the outside quote is interpreted as
+                // an argument delimiter
```
cesarbs: nit: put comments inside blocks, remove blank lines between `if`/`else if`/`else` blocks.

src/Shared/ArgumentEscaper.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                // Output any consumed backslashes and the character
+                else
+                {
+                    sb.Append('\\', backslashCount);
+                    sb.Append(arg[i]);
+                }
+            }
+
+            if (needsQuotes) sb.Append("\"");
```
cesarbs: Add `{ block }`

src/Shared/ArgumentEscaper.cs
```diff
@@ -0,0 +1,128 @@
...
+                return false;
+            }
+
+            // Only quote if whitespace exists in the string
+            return ArgumentContainsWhitespace(argument);
+        }
+
+        internal static bool IsSurroundedWithQuotes(string argument)
+        {
+            return argument.StartsWith("\"", StringComparison.Ordinal) &&
```
cesarbs: Is there anything particular to `\` that you need `StringComparison.Orginal` here? Since this is single char check you could just index the string and check what's there.

src/Shared/ArgumentEscaper.cs
```diff
@@ -0,0 +1,128 @@
...
+                   argument.EndsWith("\"", StringComparison.Ordinal);
+        }
+
+        internal static bool ArgumentContainsWhitespace(string argument)
+        {
+            return argument.Contains(" ") || argument.Contains("\t") || argument.Contains("\n");
```
cesarbs: Use `IndexOfAny(new[] { ' ', '\t', '\n' }) `.

src/Shared/DotNetMuxer.cs
```diff
@@ -0,0 +1,59 @@
...
+                return null;
+            }
+
+            var muxer = Path.Combine(muxerDir.FullName, fileName);
+            if (!File.Exists(muxer))
+            {
+                return null;
+            }
+
+            return muxer;
```
cesarbs: You can shorten these last lines with a ternary expression.

src/Shared/StringExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            => "\x1B[35m" + text + "\x1B[39m";
+
+        public static string Cyan(this string text)
+            => "\x1B[36m" + text + "\x1B[39m";
+
+        public static string White(this string text)
+            => "\x1B[37m" + text + "\x1B[39m";
+
+        public static string Bold(this string text)
+            => "\x1B[1m" + text + "\x1B[22m";
```
cesarbs: Check if we already have something like this in logging.

test/Microsoft.DotNet.Watcher.Tools.Tests/MsBuildFileSetFactoryTest.cs
```diff
@@ -264,7 +264,10 @@ public MsBuildFileSetFactoryTest(ITestOutputHelper output)
 
             var fileset = await GetFileSet(filesetFactory);
 
-            _logger.LogInformation(output.Current.GetAllLines("Sink output: "));
+            _logger.LogInformation(string.Join(
+                Environment.NewLine,
+                output.Current.Lines.Select(l => "Sink output: " + l)
+                ));
```
cesarbs: nit: move `));` to previous line

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildProjectFinder.cs
```diff
@@ -35,20 +34,20 @@ public static string FindMsBuildProject(string searchBase, string project)
 
                 if (projects.Count > 1)
                 {
-                    throw new GracefulException(Resources.FormatError_MultipleProjectsFound(projectPath));
+                    throw new FileNotFoundException(Resources.FormatError_MultipleProjectsFound(projectPath));
```
cesarbs: I was going to complain about throwing `FileNotFoundException` when in fact you've found multiple instances of what you were looking for, but I see it makes things easier for the consuming code 😄

.gitignore
natemcmaster: Project Rider :)

src/Microsoft.DotNet.Watcher.Tools/DotNetWatcher.cs
natemcmaster: The escaping has to account for backslashes, spaces, and quotes. cref https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
natemcmaster: I didn't copy this from `Command.CreateDotNet()`. The reason for dropping this is that their Command API is really designed for the CLI's internal usage, and has layers of logic to resolve commands listed in the .csproj. We don't really need that here. We just want to run a process and capture its output.

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
natemcmaster: It's a personal preference: I don't like mixing interpolation and using the string builder. But doesn't really matter either way

src/Shared/StringExtensions.cs
natemcmaster: FWIW the next step on my list is removing these and cleaning up logging altogether, as per the issue you opened #227

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildProjectFinder.cs
natemcmaster: Seemed like the most suitable type. But we could always create our own type, i.e. `ProjectNotFoundException`

PR #234, opened 2016-11-21
Abstract the console

natemcmaster:
Some projects in this repo were already using `IConsole`. This unifies the implementations to all use it.

cc @cesarbs


src/Microsoft.DotNet.Watcher.Tools/Internal/FileSetWatcher.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Tools.Internal;
```
muratg: Is this required here? (And a few other files below)

natemcmaster: Yes. I moved an API usage in this (and the other files) into a different namespace

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/project.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "dotnet-test-xunit": "2.2.0-preview2-build1029",
     "Microsoft.DotNet.Cli.Utils": "1.0.0-preview3-004056",
-    "Microsoft.DotNet.InternalAbstractions": "1.0.0",
+    "Microsoft.DotNet.InternalAbstractions": "1.0.500-preview2-1-003177",
```
muratg: When are we going to revert this? (Or how will we remember to revert it?)  I guess `-*` doesn't work?

natemcmaster: This will go away when we convert to VSTest. The issue seems to be that dotnet-test-xunit and Microsoft.DotNet.Cli.Utils have conflicting dependencies. When we switch to xunit.runner.visualstudio, this conflict will go away.

PR #236, opened 2016-11-21
Remove dependency on Microsoft.DotNet.Cli.Utils from dotnet-user-secrets

natemcmaster:
We have a cyclical dependency with CLI. We don't really need this API so this removes all usages of API from Microsoft.DotNet.Cli.Utils.

cref https://github.com/dotnet/cli/issues/4745

cc @muratg 
cc @Eilon - plus this removes all usages of the "GracefulException"  :)

### DIFF
--- a/src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
+++ b/src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
@@ -3,11 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using Microsoft.DotNet.Cli.Utils;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Tools.Internal;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Internal
 {
@@ -24,7 +25,7 @@ namespace Microsoft.Extensions.SecretManager.Tools.Internal
             _logger = logger;
         }
 
-        public string Resolve(string project, string configuration = Constants.DefaultConfiguration)
+        public string Resolve(string project, string configuration = "Debug")
         {
             var finder = new MsBuildProjectFinder(_workingDirectory);
             var projectFile = finder.FindMsBuildProject(project);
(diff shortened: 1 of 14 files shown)


muratg:
I'd like @Eilon to take a look at the new class names (*Strategy), but apart from that this looks good to me 👍

src/Microsoft.Extensions.SecretManager.Tools/Internal/SetCommand.cs
```diff
@@ -34,72 +31,74 @@ public static void Configure(CommandLineApplication command, CommandLineOptions
...
-            {
-                SetFromArguments(context);
-            }
+                options.Command = new ForOneValueStrategy(nameArg.Value, valueArg.Value);
+                return 0;
+            });
         }
 
-        private void ReadFromInput(CommandContext context)
+        public class FromStdInStrategy : ICommand
```
natemcmaster: Open to naming suggesions. Alternatives I considered:

ConsoleInputSetCommand
ConsoleInputStrategy
ICanHazPipedInputCommand

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -35,32 +36,37 @@ public string Resolve(string project, string configuration = Constants.DefaultCo
...
-            if (commandResult.ExitCode != 0)
+            var args = new[]
             {
-                _logger.LogDebug(string.Join(Environment.NewLine, commandOutput));
-                throw new GracefulException(Resources.FormatError_ProjectFailedToLoad(projectFile));
+                "msbuild",
+                targetFile,
+                "/nologo",
+                "/t:_FindUserSecretsProperty",
+                $"/p:Project={projectFile}",
```
Eilon: This is just moved code (not new), but is there a bug here if there are spaces in anything? Are spaces valid in any of these values? (Should log this as a separate bug, assuming it's actually a bug.)

Eilon: (Ditto for the following line(s).

Eilon:
The "strategy" naming strategy seems fine to me. (Don't super care either way 😄 )

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
natemcmaster: Spaces are handled by the call to `ArgumentEscaper.EscapeAndConcatenate` below. It will ensure args w/ spaces are escaped with quotes.

muratg:
:shipit:

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
Eilon: Is the correct way to escape arguments like this: `foo.exe "/p:foo bar" "/p:baz qux"`

Or like this? `foo.exe /p:"foo bar" /p:"baz qux"`

Because the former looks kind of wrong to me - though I would imagine it depends on how any particular command line tool actually works.

Does this actually work with MSBuild?

natemcmaster: Actually, both are valid. The escaper does `"/p:Prop=Value With Spaces"` so that we don't need to know about MSBuild's switch syntax.

PR #237, opened 2016-11-23
Create and use console reporter API

natemcmaster:
The reporter API breaks down complex formatter in to composable compontents. Adds 'ReporterBuilder' and 'FormatterBuilder' as syntax sugar for creating reporters with complex formatting rules

Changes:
 - Remove dependency on Microsoft.Extensions.Logging and instead use IReporter for console output
 - Only use color output when stdout and stderr are not being redirected
 - dotnet-watch: Make the default output less noisy

Resolves https://github.com/aspnet/DotNetTools/issues/227

cc @muratg


natemcmaster:
Should resolve #227.

The formatting:
verbose = dark gray, only appears when `--verbose` specified
output = normal, white text. Does not appear if `--quiet` specified.
warn = yellow text
error = red text

dotnet-watch prefixes all its lines with `watch :` to distinguish from the sub-process's output.

Some examples of output:
`dotnet watch run`
image
`dotnet watch --verbose run`
image

`dotnet user-secrets list --verbose`
image

`dotnet user-secret list --verbose` with errors
image

cc @cesarbs

src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -143,15 +141,18 @@ private void EnsureInitialized()
...
             if (!File.Exists(projectExtensionFile))
             {
                 // ensure obj folder is available
                 Directory.CreateDirectory(Path.GetDirectoryName(projectExtensionFile));
 
                 using (var fileStream = new FileStream(projectExtensionFile, FileMode.Create))
-                using (var assemblyStream = GetType().GetTypeInfo().Assembly.GetManifestResourceStream("dotnetwatch.targets"))
+                using (
+                    var assemblyStream =
+                        GetType().GetTypeInfo().Assembly.GetManifestResourceStream("dotnetwatch.targets"))
```
cesarbs: nit: actually prefer the previous format. At least move the `var` line inline with `using`.

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -15,23 +15,32 @@ public class Program
...
         public int Run(string[] args)
         {
+            DebugHelper.HandleDebugSwitch(ref args);
+
+            _reporter = new ReporterBuilder()
+                .WithConsole(_console)
+                .Verbose(f => f.WithColor(ConsoleColor.DarkGray))
+                .Warn(f => f.WithColor(ConsoleColor.Yellow))
+                .Error(f => f.WithColor(ConsoleColor.Red))
+                .Build();
```
cesarbs: Shouldn't this use the same logic used in `CreateReporter` in the watcher?

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -86,7 +95,7 @@ public int Run(string[] args)
             }
             catch (Exception exception)
             {
-                _logger.LogCritical("An error occurred. {ErrorMessage}", exception.Message);
+                _reporter.Error("An error occurred. " + exception.Message);
```
cesarbs: nit: use string interpolation instead of `+`

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -128,13 +137,13 @@ private int CreateTableAndIndexes()
...
+                        _reporter.Output("Table and index were created successfully.");
                     }
                     catch (Exception ex)
                     {
-                        _logger.LogError(
-                            "An error occurred while trying to create the table and index. {ErrorMessage}",
-                            ex.Message);
+                        _reporter.Error(
+                            $"An error occurred while trying to create the table and index. {ex.Message}"
+                            );
```
cesarbs: nit: move `);` to line above

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
```diff
@@ -78,36 +69,62 @@ internal int RunInternal(params string[] args)
...
+        {
+            return new ReporterBuilder()
+                .WithConsole(_console)
+                .Verbose(f =>
+                {
+                    f.When(() => verbose);
+                    if (!_console.IsOutputRedirected)
+                    {
+                        f.WithColor(ConsoleColor.DarkGray);
+                    }
```
cesarbs: What if it is redirected?

src/Shared/CliContext.cs
```diff
@@ -0,0 +1,21 @@
...
+    public static class CliContext
+    {
+        /// <summary>
+        /// dotnet --verbose subcommand
+        /// </summary>
+        /// <returns></returns>
+        public static bool IsGlobalVerbose()
+        {
+            bool globalVerbose;
+            bool.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_CONTEXT_VERBOSE"), out globalVerbose);
```
cesarbs: A lot of environment variables work with `true` and `1`. Shouldn't you check that here? Or do the docs instruct people to only set this to `true` if they want to enable it?

src/Shared/Reporting/ColorFormatter.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Tools.Internal
+{
+    public class ColorFormatter : IFormatter
```
cesarbs: Don't we have similar logic in the console logger already?

src/Shared/Reporting/FormattingReporter.cs
```diff
@@ -0,0 +1,62 @@
...
+        public void Output(string message)
+            => Write(_console.Out, _output.Format(message));
+
+        public void Warn(string message)
+            => Write(_console.Out, _warn.Format(message));
+
+        public void Error(string message)
+            => Write(_console.Error, _error.Format(message));
+
+        private void Write(TextWriter writer, string message)
```
cesarbs: You're not using `writer` here.

cesarbs: Add tests to verify each of the methods that call this are writing to the expected stream.

natemcmaster: Oh wow, yeah, dumb mistake.

src/Shared/Reporting/ColorFormatter.cs
natemcmaster: Yeah. It has an `IConsole` API internally. The API was insufficient because it didn't provide access to `Console.In`. 

Follow up: we remove duplication by putting this API and others in a common location to be used by Logging.Console and here as well? Could help with https://github.com/aspnet/Logging/issues/491

src/Shared/CliContext.cs
natemcmaster: it's always the result of `bool.ToString()`. cref https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/Program.cs#L164

src/Microsoft.Extensions.SecretManager.Tools/Program.cs
natemcmaster: It's output is colorless. Users might pipe commands to a file. We want to avoid spitting out the ANSI color codes.

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
natemcmaster: This tool doesn't actually have `--verbose` flags. Though I suppose detecting output redirection for colorless output is still valid...although probably a very unlikely usage.

src/Shared/Reporting/ColorFormatter.cs
cesarbs: Yeah it would be good to consolidate these things in a single place. Can you open an issue to track this?

natemcmaster: https://github.com/aspnet/Common/issues/173

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Program.cs
```diff
@@ -34,17 +33,10 @@ public int Run(string[] args)
...
-                .Verbose(f => f.WithColor(ConsoleColor.DarkGray))
-                .Warn(f => f.WithColor(ConsoleColor.Yellow))
-                .Error(f => f.WithColor(ConsoleColor.Red))
-                .Build();
-
             try
             {
                 var description = "Creates table and indexes in Microsoft SQL Server database " +
-                    "to be used for distributed caching";
+                                  "to be used for distributed caching";
```
cesarbs: nit: make it a single string.

cesarbs:
Just approved, but: are you adding the tests to make sure each level is writing to the expected stream (out or err)?

test/Microsoft.Extensions.Tools.Tests/ReporterTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        [Fact]
+        public void SkipsColorCodesForEmptyOrNullInput()
+        {
+            var formatter = new ColorFormatter(ConsoleColor.Blue);
+            Assert.Empty(formatter.Format(string.Empty));
+            Assert.Null(formatter.Format(null));
+        }
+
+        [Fact]
+        public void WritesToStandardStreams()
```
natemcmaster: @cesarbs Added it :)

cesarbs: :shipit: when tests pass.

PR #238, opened 2016-11-24
Build with MSBuild

### LINKED ISSUE #179: Remove custom packaging logic
Revert https://github.com/aspnet/DotNetTools/pull/174 when dotnet-cli supports package types.

natemcmaster:
Please take a look. The migration of project.json to csproj is mostly done. The shared infrastructure around it needs work (e.g. KoreBuild's replacement).

Resolves #179 

cc @bricelam


build.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <Target Name="Pack">
+        <MSBuild Targets="Rebuild;Pack"
+                 Projects="@(SourceProjects)"
+                 Properties="PackageOutputPath=$(ArtifactsPath)" />
+    </Target>
+
+    <Target Name="Test">
+        <MSBuild Targets="VSTest"
+                 Projects="@(TestProjects)" />
```
bricelam: I think this will run test assemblies in parallel. You may want to use task batching (i.e. `Projects="%(TestProjects.Identity)"`) to prevent this.

build.ps1
```diff
@@ -1,67 +1,26 @@
...
+    & $env:DOTNET_HOME/dotnet-install.ps1 -InstallDir $env:DOTNET_HOME -Version $env:DotnetCliVersion
 }
 
-&"$buildFile" $args
\ No newline at end of file
+& $env:DOTNET_HOME/dotnet.exe restore build.xml
+if ($LASTEXITCODE -ne 0) {
+    throw 'Restoring packages for build.xml failed'
+}
+& $env:DOTNET_HOME/dotnet.exe msbuild build.xml /nologo /v:m $args
```
bricelam: Add `/m` for a bit more oomph.

build.xml
bricelam: Oh, I lied. `BuildInParallel` defaults to `false`. But in that case, you may want to consider adding it wherever you use the `MSBuild` task with multiple projects. It does a topological sort and builds the parts in can in parallel.

build.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <Target Name="Clean">
+        <RemoveDir Directories="$(ArtifactsPath);$(MSBuildThisFileDirectory)testWorkDir" />
+        <MSBuild Projects="$(SolutionFile)" Targets="Clean" />
+    </Target>
+
+    <Target Name="Pack">
+        <MSBuild Targets="Rebuild;Pack"
+                 Projects="@(SourceProjects)"
+                 Properties="PackageOutputPath=$(ArtifactsPath)" />
```
bricelam: May also want to flow `Configuration=$(Configuration)`

bricelam: (On other `MSBuild` usages too.)

build.ps1
natemcmaster: ooh, nice. Yeah, let's put that CPU to work

build.xml
natemcmaster: So i'm still a little fuzzy on how properties flow form outer to inner builds, but I thought properties like `Configuration` just automatically flow into inner builds unless explicitly listed in 'RemoveProperties'...

build.xml
```diff
@@ -0,0 +1,52 @@
...
+        <MSBuild Projects="$(SolutionFile)" Targets="Restore" />
+    </Target>
+
+    <Target Name="Clean">
+        <RemoveDir Directories="$(ArtifactsPath);$(MSBuildThisFileDirectory)testWorkDir" />
+        <MSBuild Projects="$(SolutionFile)" Targets="Clean" />
+    </Target>
+
+    <Target Name="Pack">
+        <MSBuild Targets="Rebuild;Pack"
```
natemcmaster: Good point. Will remove.

build.xml
natemcmaster: 👍 Would be good if it works well. I'll play it. Not sure how well VSTest handles parallel test projects.

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
+  <Import Project="..\..\build\common.props" />
```
davidfowl: ..........

build.xml
```diff
@@ -0,0 +1,52 @@
...
+    </ItemGroup>
+
+    <!-- build settings -->
+    <PropertyGroup>
+        <ArtifactsPath>$(MSBuildThisFileDirectory)artifacts\build\</ArtifactsPath>
+        <SolutionFile>$(MSBuildThisFileDirectory)DotNetTools.sln</SolutionFile>
+    </PropertyGroup>
+    <ItemGroup>
+        <SourceProjects Include="src\*\*.csproj" />
+        <TestProjects Include="test\*\*.csproj" />
```
natemcmaster: Would be awesome. My MSBuild-ese isn't quite good enough to know the best way to code a good sorting algorithm like that....does the .NET Core MSBuild support code tasks?

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
davidfowl: Remind me, by default this is relative to the project file right? I still think this should be a package.

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.0.0" />
+    <PackageReference Include="Microsoft.NET.Sdk" Version="1.0.0-*" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="1.0.0" />
+
+    <!-- Restore this when https://github.com/dotnet/sdk/issues/115 is resolved -->
+    <!--<PackageReference Include="Microsoft.Extensions.Process.Sources" Version="1.1.0-*" PrivateAssets="All" />-->
+
+    <!-- ExcludeAssets="build" to workaround https://github.com/dotnet/corefx/issues/12913 -->
+    <PackageReference Include="Microsoft.DiaSymReader.Native" Version="1.4.0" ExcludeAssets="build" PrivateAssets="All" />
```
davidfowl: What is this for?

build.xml
bricelam: I don't think we should test in parallel. Just build/pack.

build.xml
bricelam: > does the .NET Core MSBuild support code tasks

Not last I checked.

natemcmaster: 😢

build.xml
bricelam: Calling the MSBuild task creates a new context--no properties are preserved unless you pass them here.

bricelam: AFAIK, the `RemoveProperties` is used to remove properties specified in the `Projects` items' `Properties` and `AdditionalProperties` metadata.

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
natemcmaster: > by default this is relative to the project file right?

Yes.

> I still think this should be a package.

Agreed. As per ongoing discussion in https://github.com/aspnet/specs/pull/49, I fully intend to move most of these common properties to a package.  When that happens, we can remove duplication from that file. That said, there is still a high chance we will still want a common.props file per repo for settings that truly are repo specific, such as `RepositoryUrl`.

bricelam: > this is relative to the project file right

Yes, the project you pass to `msbuild`, and you use `$(MSBuildThisFileDirectory)` to reference files relative to the current one.

build.xml
bricelam:
```XML
<ItemGroup>
  <Project Include="SomeProject A='A' />
</ItemGroup>

<MSBuild Projects="@(Project)" RemoveProperties="A" />
```

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
natemcmaster: It was a workaround for https://github.com/dotnet/roslyn/issues/15137. Just checked again and the workaround can be removed now.

build.xml
bricelam: I ❤️ MSBuild. (In a Stockholm syndrome kind of way)

build.xml
bricelam: I'd just do it in two passes.

```XML
<ItemGroup>
  <UnitTestProject Include="test/**/*.Test.csproj" />
  <OtherTestProject Include="test/**/*.csproj" Exclude="@(UnitTestProject)" />
</ItemGroup>
```

natemcmaster: Yeah, that's easier. added it.

natemcmaster:
:watch: Putting this work on hold. Found some un-workaround-able issues. 

https://github.com/Microsoft/vstest/issues/241 - dotnet test returns exit code 0 even when tests fail
https://github.com/Microsoft/vstest/issues/254 - VSTest breaks TeamCity test reporting

build.xml
```diff
@@ -0,0 +1,87 @@
...
+    <Import Project="build\common.props" />
+
+    <!-- build project dependencies -->
+    <PropertyGroup>
+        <RestoreSources>$(RestoreSources);https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json</RestoreSources>
+        <!-- this build project targets netcoreapp1.0 so it can restore build targets, but doesn't actually compile sources -->
+        <TargetFramework>netcoreapp1.0</TargetFramework>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="NuGetPackageVerifier" Version="1.0.1-*" />
```
ajaybhargavb: Version 1.0.2-*?

natemcmaster:
:up: :date: Changing course. Sticking with KoreBuild for now to simplify our conversion process.

Removing wip: label. This PR is ready.

cc @prafullbhosale can you take a look?

build/common.props
```diff
@@ -0,0 +1,42 @@
...
+
+        <!-- build options -->
+        <WarningsAsErrors>true</WarningsAsErrors>
+        <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Key.snk</AssemblyOriginatorKeyFile>
+        <SignAssembly>true</SignAssembly>
+        <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="$(PackWithoutBuildNumber)">
+        <PackageVersion>$(VersionPrefix)</PackageVersion>
```
natemcmaster: In this repo, version.props has it...but you make a good point. This should be made safer so that it can be generalized.

src/Microsoft.Extensions.Caching.SqlConfig.Tools/Properties/AssemblyInfo.cs
```diff
@@ -2,10 +2,5 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Reflection;
-using System.Resources;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
```
natemcmaster: yes, still required. This is different from the serviceable tag on the nupkg.

build/common.props
prafullbhosale: version.props only sets the `VersionPrefix` and `VersionSuffix` but not the `PackageVersion`. 
Am I missing something?

natemcmaster: Yeah, PackageVersion is in NuGet's pack targets. It defaults to `$(Version)` but can be overridden if you want the pkg version to differ from the build version.

prafullbhosale: 👍 Did not know that.

PR #241, opened 2016-12-07
Update packaging for tools

natemcmaster:
Downgrade to Microsoft.NETCore.App 1.0.0. (cref https://github.com/aspnet/Coherence/issues/141)
Add 'prefercliruntime' to package. (cref https://github.com/dotnet/cli/pull/4814).
Move targets files in to 'toolassets'. The 'tools' folder has conventions in NuGet that we don't want applied to these packages.

cc @muratg @prafullbhosale


src/Microsoft.Extensions.Caching.SqlConfig.Tools/Microsoft.Extensions.Caching.SqlConfig.Tools.nuspec
```diff
@@ -20,6 +20,7 @@
     </dependencies>
   </metadata>
   <files>
+    <file src="prefercliruntime" target="\prefercliruntime" />
```
prafullbhosale: Does the leading `\` have any significance?

natemcmaster: I don't think so

PR #243, opened 2016-12-07
Use MSBuild project extensions instead of importing the users project

### LINKED ISSUE #242: dotnet-user-secrets breaks when project uses implicit imports
Repro:
```xml
<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp1.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NETCore.App" Version="1.0.1" />
    <DotNetCliToolReference Include="Microsoft.Extensions.SecretManager.Tools" Version="1.0.0-msbuild2-final" />
  </ItemGroup>
</Project>
```
dotnet-cli 1.0.0-preview3-004215
Run any `dotnet user-secrets` command

Error:
```
debug: Project file path C:\Users\namc\Documents\Visual Studio 2017\Projects\WebApplication1\src\WebApplication2\WebApplication2.csproj.
debug: C:\Program Files\dotnet\sdk\1.0.0-preview4-004215\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.targets(35,3): error MSB4102: The value "" of the "Project" attribute in element <Import> is invalid. Parameter "path" cannot have zero length. [C:\Users\namc\.nuget\packages\microsoft.extensions.secretmanager.tools\1.0.0-msbuild2-final\tools\FindUserSecretsProperty.targets]
fail: Could not load the MSBuild project 'C:\Users\namc\Documents\Visual Studio 2017\Projects\WebApplication1\src\WebApplication2\WebApplication2.csproj'.
```

cref https://github.com/Microsoft/msbuild/issues/1392

natemcmaster:
Implicit imports prevents using <Import> on a project file that has the Sdk attribute. This change instead generates a file in the MSBuildProjectExtensionsPath to inject targets require to find the UserSecretsId property in a project. (This is the approached used by dotnet-watch and dotnet-ef).

Resolves #242 

cc @mlorbetske @muratg 
cc @bricelam can you review? This is similar to the approach you've implemented in dotnet-ef to extract project metadata.

### DIFF
--- a/src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
+++ b/src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
@@ -5,19 +5,16 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Reflection;
 using Microsoft.Extensions.Tools.Internal;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Internal
 {
-    public class ProjectIdResolver : IDisposable
+    public class ProjectIdResolver
     {
-        private const string TargetsFileName = "FindUserSecretsProperty.targets";
         private const string DefaultConfig = "Debug";
         private readonly IReporter _reporter;
         private readonly string _workingDirectory;
-        private readonly List<string> _tempFiles = new List<string>();
 
         public ProjectIdResolver(IReporter reporter, string workingDirectory)
         {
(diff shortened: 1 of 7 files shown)


natemcmaster:
cc @rainersigwald

src/Microsoft.Extensions.SecretManager.Tools/Internal/ProjectIdResolver.cs
```diff
@@ -74,42 +71,27 @@ public void Dispose()
...
-            // test project has an always-up to date version of the targets file.
-            // TODO cleanup after we switch to an MSBuild system in which can specify "CopyToOutputDirectory: Always" to resolve this issue
-            var outputPath = Path.GetTempFileName();
-            using (var resource = GetType().GetTypeInfo().Assembly.GetManifestResourceStream(TargetsFileName))
-            using (var stream = new FileStream(outputPath, FileMode.Create))
+            // relies on MSBuildProjectExtensionsPath and Microsoft.Common.targets to import this file
+            // into the target project
+            var projectExtensionsPath = Path.Combine(
+                Path.GetDirectoryName(projectFile),
+                "obj",
```
rainersigwald: This embeds some assumptions about the path to `IntermediateObjectDirectory` that won't always hold--but are probably ok for most new (or converted) .NET Core projects.

natemcmaster:
> the hardcoding of ./obj/ will have to be changed before RTM.

Good to know. Where will MSBuildProjectExtensionsPath default to by RTM?

rainersigwald:
@natemcmaster I expect that to stay the default (it's been like that for years), but the location of the obj directory is a user-specifiable option. Many teams prefer, for example, to have a single `{repoRoot}\out` directory that's easy to delete. Some folks even go so far as to put all build outputs on a different drive that they can `format` to get back to a clean state.

bricelam:
We [allow overriding](https://github.com/aspnet/EntityFramework.Tools/blob/feature/msbuild/src/Microsoft.EntityFrameworkCore.Tools.DotNet/Program.cs#L45-L47) it in dotnet-ef, but it does feel like we're abusing this feature a bit... Is there a way outside of MSBuild to get the value of the `MSBuildProjectExtensionsPath` property?

natemcmaster:
Ok. Do you have any ideas how a tool like this could evaluate the value of MSBuildProjectExtensionsPath?

bricelam:
Maybe, in general, we just need a built-in target to get the value of properties. E.g. `msbuild /t:GetPropertyValue /p:Property=UserSecretsId /verbosity:minimal`

rainersigwald:
You could use the OM to extract the property, but then you could just get the property you wanted instead (which is what I want to talk to you about after resolving the immediate issue).

For this limited case, you could probably force-override `MSBuildProjectExtensionsPath` with a command-line property to a folder you control (and clean up after execution). That might hork the project build in other ways, but since you're only calling one target that you control it might be ok?

natemcmaster:
OM? Sorry, not familiar with that abbreviation.

Overriding the MSBuildProjectExtensionsPath prop would _probably_ work for dotnet-user-secrets, but quickly falls over for anything that needs properties that depend on other project extensions such as those generated by NuGet restore.
 
Let's continue this conversation on https://github.com/aspnet/DotNetTools/issues/244.

rainersigwald:
OM is "Object model"--I'm saying you could just evaluate the project programmatically and ask MSBuild what the property you want is.

natemcmaster:
Gotcha. Unfortunately, we went down that path and got burned too many times by MSBuild API. At the strong recommendation of David Kean and others, we have been pursuing project-evaluation-as-a-target instead.

natemcmaster:
Retargeted to feature/msbuild. This fix won't make it into RC.2, but will be in the next update.

PR #256, opened 2017-01-20
Upgrade to RC.3

natemcmaster:
Removes default globs, temporary sources, and uses implicit framework references


BrennanConroy:
:shipit:

PR #262, opened 2017-02-10
Cleanup project files and workaround, run tests on .NET Core 1.1

natemcmaster:
Remove workarounds and TODOS that crept in during MSBuild conversion.

Also, update tests to run on .NET Core 1.1. Now that .NET Core CLI runs on .NET Core 1.1 itself, this is more reflective of how the CLI tools will execute in a user project.


test/Microsoft.DotNet.Watcher.Tools.Tests/MsBuildFileSetFactoryTest.cs
```diff
@@ -28,11 +28,9 @@ public MsBuildFileSetFactoryTest(ITestOutputHelper output)
         [Fact]
         public async Task FindsCustomWatchItems()
         {
-            TemporaryCSharpProject target;
             _tempDir
-                .WithCSharpProject("Project1", out target)
+                .WithCSharpProject("Project1", out var target)
```
cesarbs: 👍

PR #263, opened 2017-02-11
1.0.4 patch train updates

JunTaoLuo:


test/Microsoft.Extensions.SecretManager.Tests/ConfigurationExtensionTests.cs
```diff
@@ -18,7 +19,7 @@ public void AddUserSecrets_Does_Not_Fail_On_Non_Existing_File_Explicitly_Passed(
                                 .AddUserSecrets(userSecretsId: Guid.NewGuid().ToString());
         }
 
-        [Fact]
+        [ConditionalFact (Skip = "Obsolete API")]
```
natemcmaster: You don't need ConditionalFact to do this. `[Fact(Skip = "reason")]`.

Also, don't skip these tests. They should still work even though API is obsolete.

test/Microsoft.Extensions.SecretManager.Tests/project.json
```diff
@@ -1,11 +1,15 @@
 {
   "buildOptions": {
     "warningsAsErrors": true,
+    "nowarn": [
+      "CS0618" // Tests using obsolete APIs are skipped.
```
natemcmaster: Don't add a project-wide nowarn. Use `#pragma` to disable it. Example: 
image

test/Microsoft.Extensions.SecretManager.Tests/project.json
```diff
@@ -1,11 +1,15 @@
...
+    ],
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
     "dotnet-test-xunit": "1.0.0-rc3-000000-01",
-    "Microsoft.Extensions.SecretManager.Tools": "1.0.0-preview4-final",
+    "Microsoft.AspNetCore.Testing": "1.0.1",
```
natemcmaster: We don't need to add a new dependency. ConditionalFacts not required.

test/Microsoft.Extensions.SecretManager.Tests/ConfigurationExtensionTests.cs
JunTaoLuo: Ahh

JunTaoLuo: Actually I don't think there is much value keeping these tests around. Why not just remove them?

test/Microsoft.Extensions.SecretManager.Tests/ConfigurationExtensionTests.cs
```diff
@@ -19,37 +19,6 @@ public void AddUserSecrets_Does_Not_Fail_On_Non_Existing_File_Explicitly_Passed(
         }
 
         [Fact]
-        public void AddUserSecrets_Does_Not_Fail_On_Non_Existing_File()
```
natemcmaster: Please keep these. Even though we've obsoleted the API, we want to ensure we haven't regressed on the tool.

JunTaoLuo:
No longer building this repo in 1.0.4 patch train.

PR #269, opened 2017-03-10
Set dotnet-watch env variable on inner process

### LINKED ISSUE #52: Make it possible to detect when a process is running inside dotnet-watch
When using approvaltests or similar it will pop up a diff tool when the test does not match the 'gold' template.

Is there a way to know if running inside dnx-watch? If not would you consider adding an environmental variable called `DNX_WATCH=1` (or `DOTNET_WATCH=1` for rc2 so you can know if you are running inside DNX watch?

natemcmaster:
Makes it possible to detect when running inside dotnet-watch by setting 'DOTNET_WATCH=1'

Resolves #52

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Tools/ProcessSpec.cs
+++ b/src/Microsoft.DotNet.Watcher.Tools/ProcessSpec.cs
@@ -11,6 +11,7 @@ namespace Microsoft.DotNet.Watcher
     {
         public string Executable { get; set; }
         public string WorkingDirectory { get; set; }
+        public IDictionary<string, string> EnvironmentVariables { get; } = new Dictionary<string, string>();
         public IEnumerable<string> Arguments { get; set; }
         public OutputCapture OutputCapture { get; set; }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Watcher.Tools/ProcessSpec.cs
```diff
@@ -11,6 +11,8 @@ public class ProcessSpec
     {
         public string Executable { get; set; }
         public string WorkingDirectory { get; set; }
+        public IDictionary<string, string> EnvironmentVariables { get; }
+            = new Dictionary<string, string>();
```
cesarbs: nit: leave it on the same line

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/TestProjects/KitchenSink/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
cesarbs: Add copyright header.

natemcmaster: Done

natemcmaster:
rebased. :watch: on ci passes.

PR #272, opened 2017-03-14
Update dependencies and build version to 1.2.0

natemcmaster:
- Unify dependencies.
- Remove duplicated code from aspnet/common. This duplicated code was added because of the weird timing and forking required to build the 1.0.0 tools. We can now use a package reference.
- Update appveyor image

cc @Eilon 

@JunTaoLuo  can you take a look? This is just build stuff, no functional changes here.


src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
```diff
@@ -20,7 +20,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="$(AspNetCoreVersion)" />
```
Eilon: Very late to the party here, but why is this changed to use the package instead of shared source? Or am I reading this wrong? (I'm not sure it matters much, but I'm curious about the change.)

natemcmaster: The 1.1.0-preview4-final (project.json) version of tools use API from M.Ext.CommandLineUtils that was in the 1.1.0 release, but was missing from 1.0.0. Because the MSBuild fork of this repo had to roll back from 1.1.0 dependencies to 1.0.0, we copied the source code here. We would have used the .Sources package, but the pre-release version of the MSBuild SDK we used didn't support .Source packages at the time. (Sidenote: the rel/1.0.0 branch of this project is still on an early pre-release SDK. It won't open in VS 2017

 This PR upgrades to 2.0 dependencies which means we can revert to using a package reference.

Eilon: Ah, such a complex world, thank you for the explanation 😄

Issue #265, opened 2017-02-20
dotnet-watch prints invalid characters in Command Prompt


justdmitry:
`dotnet-watch` prints unknown/non-ascii characters in log messages (red boxes on image).

![cmd_2017-02-20_13-52-32](https://cloud.githubusercontent.com/assets/3540597/23122302/bb09de44-f774-11e6-8660-a7b19f95dc9a.png)

Package used: `Microsoft.DotNet.Watcher.Tools` version `1.0.0-msbuild3-final`

Environment: `win10-x64`, SDK `1.0.0-rc4-004771`, console encoding `utf-8` (green on image)

Possibly related with https://github.com/dotnet/cli/issues/2590

natemcmaster:
Can confirm. This only appears in Command Prompt, which doesn't support ANSI color codes. To workaround, use basically any other shell, like PowerShell, ConEmu, etc.

Eilon:
Will consider this for a 1.0.1 patch of this package.

Eilon:
@natemcmaster is this bug meant for the patch or for dev (2.0.0-preview1)? For dev I see you merged the change in - that's good. But then this bug should move to the correct milestone, and I guess we need a new bug opened for 1.0.1?

natemcmaster:
@Eilon I'm honestly confused about whether we have 1.0.1 planned or not. Feel free to move milestones/create duplicates as necessary to satisfy bookkeeping procedures.

snerks:
Issue appears to exist in Nightly Build for Windows 7 (SP1) 64-bit.

**Steps to reproduce:**
Follow steps provided in the blog post written by @shanselman, to create a solution, containing  a Class Library and a Unit Test library:
 [https://www.hanselman.com/blog/CommandLineUsingDotnetWatchTestForContinuousTestingWithNETCore10AndXUnitnet.aspx](https://www.hanselman.com/blog/CommandLineUsingDotnetWatchTestForContinuousTestingWithNETCore10AndXUnitnet.aspx)

Download ZIP of latest Windows X64 SDK:
https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.zip

Unzip the file. Place the contents of the unzipped `dotnet-dev-win-x64.latest` folder in the test project folder in the source code (i.e. the `mytests` folder, described in the blog post written by @shanselman) :

Open a command line (`cmd.exe`)
Navigate to the `mytests` folder

[...]

PR #273, opened 2017-03-14
[1.0.1] Fix color output in command prompt

### LINKED ISSUE #271: Fix console color output in Command Prompt
Back-port #265 to 1.0.1
cc @Eilon

natemcmaster:
Backport #268 to 1.0.1

Resolves #271 

NB: this is merging to "namc/rel/1.0.1". If approved, I'll rename that branch to "rel/1.0.1" and rebase this PR to merge into that branch.


shared/IConsole.cs
```diff
@@ -15,5 +15,7 @@ public interface IConsole
         bool IsInputRedirected { get; }
         bool IsOutputRedirected { get; }
         bool IsErrorRedirected { get; }
+        ConsoleColor ForegroundColor { get; set; }
```
Eilon: Hmm I don't recall exactly, but I think at one point we said we wanted patches to be 100% binary compatible, which means that no "public" breaking change can happen at all, even in `.Internal` namespaces. Is there a way to make this change without breaking any API surface?

natemcmaster: I don't think this guidance applies to command line tools. Users never reference any of the assemblies produced in this repo -- they are all `DotNetCliToolReference` packages -- so breaking changes shouldn't matter.

Eilon: Ah, very good point! Looks like no "regular" NuGet packages are built from this repo, so presumably no one ever references anything here as a DLL reference.

natemcmaster:
Is this approved as in merge to rel/1.0.1? Or do we still need DJ approval?

shanselman:
Is this going to happen soon?

snerks:
Is it expected that this fix will work on Windows 7 (SP1) 64-bit?

natemcmaster:
Still pending approval. It fixes Command Prompt color, which should mean Win 7 is fixed. The `dev` branch already has this fix. You can take the fix for a test drive by using our [nightly feeds](https://github.com/aspnet/home#builds).

snerks:
@natemcmaster - thanks. The steps I took were:

Download ZIP of latest Windows X64 SDK:
https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.zip

Unzip the file. Place the contents of the unzipped `dotnet-dev-win-x64.latest` folder in the test project folder in the source code (i.e. the `mytests` folder, described in the blog post written by @shanselman) :
 [https://www.hanselman.com/blog/CommandLineUsingDotnetWatchTestForContinuousTestingWithNETCore10AndXUnitnet.aspx](https://www.hanselman.com/blog/CommandLineUsingDotnetWatchTestForContinuousTestingWithNETCore10AndXUnitnet.aspx)

Open a command line (`cmd.exe`)
Navigate to the `mytests` folder

Type `dotnet --version`
`2.0.0-preview1-005660`

Type `dotnet test`

**Expected result:**
Colorised output

**Actual result:**
All output is white text

Note: I'm also seeing character-encoding issues in cmd.exe (and Powershell) but not Git-Bash, similar to the issue #265.

_It's possible I have missed some necessary steps._

natemcmaster:
@snerks I'm not sure what you're trying to do. I didn't see mention of dotnet-watch anywhere in your repro steps. This PR is only about changes to dotnet-watch, not dotnet-test. Consider [opening an issue](https://github.com/aspnet/DotNetTools/issues/new) instead of commenting on this PR if you're having trouble using our nightly builds.

snerks:
@natemcmaster - I've moved the comment to #265 and tried to clarify the steps, to show usage of `dotnet-watch`.

shanselman:
Let me know if you want to pair on this @natemcmaster. The issue is that when we do "dotnet watch SOMETHING" the colors are not used, as in "dotnet watch test"

natemcmaster:
@shanselman It sounds like there are two overlapping things being discussed. (1) the color dotnet-watch produces (2) the color dotnet-test produces when run by dotnet-watch. This PR (and issue #265) only address (1). Can you open a separate issue for (2)?

shanselman:
@natemcmaster I don't think dotnet test produces color. So I think you've fixed it with https://github.com/aspnet/DotNetTools/issues/265#issuecomment-289608589 right?

natemcmaster:
> I don't think dotnet test produces color

Not yet ;) https://github.com/Microsoft/vstest/pull/641. 

Was just making sure we were talking about the same thing. Yes, #265 is resolved. This is the only color output issue I'm aware of. If you find others, feel free to open a new issue.

shanselman:
OK, cool. We should just watch out/make sure that when it does, everything works as it should on cmd.exe and powershell.exe.

natemcmaster:
FYI 1.0.1 is now available on NuGet.org.

https://www.nuget.org/packages/Microsoft.DotNet.Watcher.Tools/1.0.1

PR #287, opened 2017-04-27
Upgrade System.Data.SqlClient to 4.4.0-*

mikeharder:
- Addresses "SqlClient fails with netcoreapp2.0 on Win7/Server2008" (https://github.com/dotnet/corefx/issues/18406)
- Revert if and when $(CoreFxVersion) is upgraded to 4.4.0-*

### DIFF
--- a/src/Microsoft.Extensions.Caching.SqlConfig.Tools/Microsoft.Extensions.Caching.SqlConfig.Tools.csproj
+++ b/src/Microsoft.Extensions.Caching.SqlConfig.Tools/Microsoft.Extensions.Caching.SqlConfig.Tools.csproj
@@ -19,7 +19,13 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.CommandLineUtils.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
-    <PackageReference Include="System.Data.SqlClient" Version="$(CoreFxVersion)" />
+
+    <!--
+      * Use 4.4.0-* instead of $(CoreFxVersion) to address "SqlClient fails with netcoreapp2.0 on Win7/Server2008"
+        * https://github.com/dotnet/corefx/issues/18406
+      * Revert if and when $(CoreFxVersion) is upgraded to 4.4.0-*
+    -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Caching.SqlConfig.Tools/Microsoft.Extensions.Caching.SqlConfig.Tools.csproj
```diff
@@ -19,7 +19,13 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.CommandLineUtils.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
-    <PackageReference Include="System.Data.SqlClient" Version="$(CoreFxVersion)" />
+
+    <!--
+      * Use 4.4.0-* instead of $(CoreFxVersion) to address "SqlClient fails with netcoreapp2.0 on Win7/Server2008"
```
natemcmaster: Please add to dependencies.props

mikeharder: This change should be temporary if we eventually update `$(CoreFxVersion)` to `4.4.0-*`, though this may or may not be happening for 2.0.  This change has already been merged like this in several other repos, and I'd like to keep this as simple as possible, so I think this is good enough for preview1.  I will create an issue to convert these to `dependencies.props` in dev.

natemcmaster: It's the only usage of `$(CoreFxVersion)` in the whole repo, so just do it now. No reason to overcomplicate this.

mikeharder: Talked in person.  We have already merged the hardcoded version to several other repos, so we will continue doing so until all repos are updated.  Then I will convert all repos to `dependencies.props` with a new variable `$(SqlClientVersion)`.

PR #291, opened 2017-04-28
1.0.1: More logging and timeouts in dotnet-watch tests

### LINKED ISSUE #290: [1.0.5 Patch][Flaky test]: Microsoft.DotNet.Watcher.Tools.FunctionalTests.AppWithDepsTests
Test was stuck for over 13 hours. Could we ensure there aren't any race conditions or checks we're missing that are causing it to be stuck? It would be nice to address this even once we get a patch build just so that it doesn't affect future patches.

natemcmaster:
Tests can hang indefinitely if dotnet-restore or dotnet-build on the test apps fails to exit.

Resolves #290

### DIFF
--- a/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
+++ b/test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
@@ -19,6 +19,8 @@ namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
     {
         private Process _process;
         private readonly ProcessSpec _spec;
+        private readonly TaskCompletionSource<string> _tcs = new TaskCompletionSource<string>();
+        private readonly CancellationTokenSource _cts = new CancellationTokenSource();
         private BufferBlock<string> _source;
         private ITestOutputHelper _logger;
         private int _reading;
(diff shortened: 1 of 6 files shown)


test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
```diff
@@ -91,28 +101,37 @@ private void StartProcessingOutput(StreamReader streamReader)
         {
             _source = _source ?? new BufferBlock<string>();
             Interlocked.Increment(ref _reading);
-            Task.Run(() =>
+            Task.Run(async () =>
             {
-                string line;
-                while ((line = streamReader.ReadLine()) != null)
+                Task<string> line;
+                while ((line = await Task.WhenAny(_tcs.Task, streamReader.ReadLineAsync())).Result != null)
```
pranavkm: Is any of this from the dev branch?

natemcmaster:
Yeah, I've gone ahead and skipped the entire test project so we can investigate this further

test/Microsoft.DotNet.Watcher.Tools.FunctionalTests/AwaitableProcess.cs
natemcmaster: No, but i'll move it to dev when we figure out if this resolved the issue

pranavkm: Gotcha. I was trying to figure out how well I need to look at the code. I'll review it 👍

natemcmaster:
Nevermind, this doesn't solve the issue. It still repros. It seems there is something fundamentally wrong with the way this code is trying to put async wrappers around stdout processing.

PR #298, opened 2017-05-22
Add task for generating ssl certificates with MSBuild

javiercn:


src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.AspNetCore.CertificateGeneration.Task
+{
+    internal static class CertificateManager
```
cesarbs: I think this should be called `CertificateGenerator`.

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
```diff
@@ -0,0 +1,110 @@
...
+            StoreName storeName,
+            StoreLocation storeLocation)
+        {
+            using (var rsa = RSA.Create(2048))
+            {
+                var signingRequest = new CertificateRequest(
+                    new X500DistinguishedName(subjectName), rsa, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+
+                var enhacedKeyUsage = new OidCollection();
+                enhacedKeyUsage.Add(new Oid("1.3.6.1.5.5.7.3.1", "Server Authentication"));
```
cesarbs: Put OID in a const.

src/Microsoft.AspNetCore.CertificateGeneration.Task/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Microsoft.AspNetCore.CertificateGeneration.Task
+{
+    public class Program
+    {
+        public static int Main(string[] args)
+        {
+            var app = new CommandLineApplication();
+            app.HelpOption("-h|--help");
+
+            app.Command("ssl", c =>
```
cesarbs: How about "self-sign" or "self-signed" instead of "ssl"? "ssl" is too generic/vague.

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
javiercn: It also finds :)

src/Microsoft.AspNetCore.CertificateGeneration.Task/Program.cs
javiercn: On the contrary, SSL is way more specific than self-sign or self-signed. self-sign or self-signed doesn't specify at all what the certificate is being used/intended for.
I can have a self-signed certificate that gets used for client authentication for example.
BTW, the consumer of this tool is the task below, so I don't think names matter that much in that regard.

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
javiercn: It's only used in one place, it's not worth it

javiercn:
@cesarbs Updated

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
```diff
@@ -0,0 +1,117 @@
...
+            IEnumerable<string> subjectAlternativeName,
+            string friendlyName,
+            DateTimeOffset notBefore,
+            DateTimeOffset expires,
+            StoreName storeName,
+            StoreLocation storeLocation)
+        {
+            using (var rsa = RSA.Create(2048))
+            {
+                var signingRequest = new CertificateRequest(
```
javiercn: @blowdart This is the cert creation part. It has basic constraints, EKUs, key usage and subject and san. Am I missing something?

src/Microsoft.AspNetCore.CertificateGeneration.Task/build/Microsoft.AspNetCore.CertificateGeneration.Task.targets
```diff
@@ -0,0 +1,17 @@
...
+
+Generates an SSL certificate to use for development in ASP.NET Core applications
+********************************************************************************************
+-->
+  <Target Name="GenerateSSLCertificate">
+    <PropertyGroup>
+      <ExecutablePath>$(MSBuildThisFileDirectory)..\tools\netcoreapp2.0\Microsoft.AspNetCore.CertificateGeneration.Task.dll</ExecutablePath>
+      <_Force Condition="$(ForceGenerateSSLCertificate) == 'true'">--force</_Force>
+    </PropertyGroup>
+    <Exec Command="dotnet $(ExecutablePath) ssl $(_Force)" />
```
natemcmaster: Two comments here:
```dotnet```
Anything that relies on PATH and is going to be fragile. Let's open an issue to track https://github.com/Microsoft/msbuild/issues/1669 for a better solution here. This is basically the same issue as https://github.com/aspnet/MvcPrecompilation/issues/50.

**MSBuild warnings/errors:**

If you don't take my suggestions above about prefixing console output, you can set `IgnoreStandardErrorWarningFormat="false"` here.

File: src/Microsoft.AspNetCore.CertificateGeneration.Task/Microsoft.AspNetCore.CertificateGeneration.Task.csproj [hunk 6]
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
```
natemcmaster: You need to add some tags to the package.

natemcmaster:
When I asked about making this an MSBuild task, you said it required API only available to .NET Core 2.0. Can you specify which APIs those were? 

I'm concerned about the maintainability of this tool as invoking a .NET Core console app from MSBuild is going to be very fragile. You cannot be sure "dotnet" is on the PATH, and if it is, you cannot be sure that the right version of Microsoft.NETCore.App is installed.

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateGenerator.cs
```diff
@@ -0,0 +1,117 @@
...
+        public static X509Certificate2 FindCertificate(string subjectValue, StoreName storeName, StoreLocation storeLocation)
+        {
+            using (var store = new X509Store(storeName, storeLocation))
+            {
+                store.Open(OpenFlags.ReadOnly);
+                var certificates = store.Certificates.Find(X509FindType.FindBySubjectDistinguishedName, subjectValue, validOnly: false);
+                var current = DateTimeOffset.UtcNow;
+
+                var found = certificates.OfType<X509Certificate2>()
+                    .Where(c => c.NotBefore <= current && c.NotAfter >= c.NotAfter && c.HasPrivateKey)
```
nil4: Should `c.NotAfter >= c.NotAfter` be `c.NotAfter >= current` instead?

src/Microsoft.AspNetCore.CertificateGeneration.Task/Microsoft.AspNetCore.CertificateGeneration.Task.csproj
javiercn: https://github.com/aspnet/DotNetTools/pull/298/files#diff-068e18b80908ccb8fbb044d9579c5c21R9

javiercn:
@natemcmaster @cesarbs Updated

src/Microsoft.AspNetCore.CertificateGeneration.Task/README.md
```diff
@@ -0,0 +1,48 @@
+﻿Microsoft.AspNetCore.CertificateGeneration.Task
+============
```
natemcmaster: Nit: make this as long as the title that it underlines.

natemcmaster:
Test failures:
**Linux:**
```
Microsoft.AspNetCore.CertificateGeneration.Task.GenerateSSLCertificateTaskTest.GenerateSSLCertificateTaskTest_CreatesCertificate_IfFoundCertificateDoesNotHavePrivateKeys
```
**macOS:** all tests.
**Windows:** package verifier doesn't like this package. You can add an exclusion to NPGV.json for the BUILD_ITEMS_FRAMEWORK warning, but we need to add MSBuild to Microsoft.Build.Tasks.Core to the list of allowed packages. https://github.com/aspnet/BuildTools/wiki/Third-party-dependency-version-check

test/Microsoft.AspNetcore.CertificateGeneration.Task.Tests/GenerateSSLCertificateTaskTest.cs
```diff
@@ -0,0 +1,226 @@
...
+                signingRequest.CertificateExtensions.Add(sanBuilder.Build());
+
+                var certificate = signingRequest.CreateSelfSigned(
+                    notBefore == default(DateTimeOffset) ? DateTimeOffset.Now : notBefore,
+                    expires == default(DateTimeOffset) ? DateTimeOffset.Now.AddYears(1) : expires);
+
+                if (savePrivateKey)
+                {
+                    var export = certificate.Export(X509ContentType.Pkcs12, "");
+                    var imported = new X509Certificate2(export, "", X509KeyStorageFlags.PersistKeySet);
```
natemcmaster: Need to mark the cert as exportable. `X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable`

test/Microsoft.AspNetcore.CertificateGeneration.Task.Tests/GenerateSSLCertificateTaskTest.cs
```diff
@@ -0,0 +1,226 @@
...
+            DateTimeOffset notBefore = default(DateTimeOffset),
+            DateTimeOffset expires = default(DateTimeOffset),
+            bool savePrivateKey = true)
+        {
+            using (var rsa = RSA.Create(2048))
+            {
+                var signingRequest = new CertificateRequest(
+                    new X500DistinguishedName(TestSubject), rsa, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+
+                var enhacedKeyUsage = new OidCollection();
```
natemcmaster: Typo: did you mean `enhancedKeyUsage`?

javiercn:
@bartonjs We are trying to save some certificates into the certificate store on OSX, but we are running into some issues https://travis-ci.org/aspnet/DotNetTools/jobs/235795184#L1614 Can you help us out?

Here are the places where we try to save the certificate:
https://github.com/aspnet/DotNetTools/pull/298/files#diff-df27b02c4c684ea931e2137c76ad631eR64
https://github.com/aspnet/DotNetTools/pull/298/files#diff-beb92a6c536a71e60b36c3dd4b16f966R190

javiercn:
@bartonjs We figured out I think. There might be an issue on OSX, I filed a bug for it https://github.com/dotnet/corefx/issues/20272

src/Microsoft.AspNetCore.CertificateGeneration.Task/CertificateManager.cs
```diff
@@ -0,0 +1,96 @@
...
+                signingRequest.CertificateExtensions.Add(new X509KeyUsageExtension(X509KeyUsageFlags.KeyEncipherment, critical: true));
+                signingRequest.CertificateExtensions.Add(
+                    new X509BasicConstraintsExtension(
+                        certificateAuthority: false,
+                        hasPathLengthConstraint: false,
+                        pathLengthConstraint: 0,
+                        critical: true));
+
+                var sanBuilder = new SubjectAlternativeNameBuilder();
+                foreach (var alternativeName in subjectAlternativeName)
```
blowdart: One addition I'd do here, simply because I did it in our internal CA, is, if the subject name is not in the san names and there are san names, then put it in the san. The specs say if there's a SAN then ignore the subject, but people don't realise that. Note that this means, strictly speaking, subject name is optional.

natemcmaster: @blowdart as this PR has been merged already, should we open a new issue to fix this before preview2?

blowdart: No, it's a nice to have. In Javier's usage it doesn't matter.

PR #340, opened 2017-08-25
Use PackageLineup to manage PackageReference versions

natemcmaster:
Removes version info from this repo in favor of lineups.


NuGet.config
```diff
@@ -3,7 +3,6 @@
   <packageSources>
     <clear />
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
-    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
```
ryanbrandenburg: Is this just incidental, or is it possible due to some other change?

testWorkDir/Directory.Build.props
```diff
@@ -0,0 +1,3 @@
+<Project>
```
ryanbrandenburg: This is something we'll have to do in all of our sample-like stuff right?

natemcmaster: We don't need to do it in samples in general. This is only required for dotnet-watch's tests because it is generating csproj files on the fly for its functional test.

NuGet.config
natemcmaster: Intentional. As a part of lineups, we snap Internal.AspNet.Sdk and API check PackageReferences to the version bundled in KoreBuild.

cref https://github.com/aspnet/BuildTools/blob/03a29fe5f7fb1eace6ae85351bf708ca8b851144/modules/BundledPackages/module.targets#L4

PR #341, opened 2017-08-28
Use M.E.Process.Sources 2.0.0

JunTaoLuo:
I think this goes directly into patch/1.0.6

### DIFF
--- a/src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
+++ b/src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
@@ -20,7 +20,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <!-- Using 2.0.0 instead of 1.0.x since it does not use contentFiles  -->
+    <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="2.0.0" PrivateAssets="All" />
   </ItemGroup>
 
 </Project>


src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="1.0.1" PrivateAssets="All" />
```
pranavkm: cc @natemcmaster. Pinning it to a patch appropriate version.

pranavkm:
:shipit:

src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj
natemcmaster: Does this compile? IIRC this version of the package doesn't work because it doesn't use contentFiles.

JunTaoLuo: Ah... that's why this was odd. I guess we should pin it to 2.0.0 then? Is it okay that these tools depend on 2.0.0 packages.

natemcmaster: Yeah, this repo branched and built at a weird time. It was out of band with the runtime shipping mechanisms. This looks like a weird case where we missed something during pinning.

I think 2.0.0 should be fine. The one c# fine in this package hasn't changed in a long time.

JunTaoLuo:
Yea this now builds correctly.

Issue #244, opened 2016-12-07
Determine value of  MSBuildProjectExtensionsPath 


natemcmaster:
dotnet-watch and dotnet-user-secrets assume `MSBuildProjectExtensionsPath` is the default value is `obj\`, but users can change this by moving their intermediate path. We should attempt to discover its actual value. This path is required to be correct in order to find and import generated targets files.

cref https://github.com/aspnet/DotNetTools/pull/243#issuecomment-265599129

See https://github.com/Microsoft/msbuild/blob/052ad30f8f706a87e9fa9431fda1ec9c0b73893b/src/XMakeTasks/Microsoft.Common.targets#L127

**Workarounds**
1. Users can manually import the generated targets from their csproj.
```xml
<Import Project="obj\$(MSBuildProjectFile).dotnetwatch.g.targets" Condition="Exists('obj\$(MSBuildProjectFile).dotnetwatch.g.targets')" />
```
2. Pass in the path explicitly so the targets are generated in a different folder. (requires 2.0.0 preview2 or newer)
```
dotnet watch --msbuildprojectextensionspath build/obj run
```

natemcmaster:
Might be possible if CLI were to implement this feature: https://github.com/dotnet/cli/issues/5300. Otherwise, I'm not quite sure how we implement this correctly without opening ourselves to all the problems of using MSBuild API directly.

muratg:
Depends on dotnet/cli#5300.

natemcmaster:
Another workaround idea: we could write the dotnet-watch targets to MSBuildUserExtensionsPath instead of MSBuildProjectExtensionsPath.

natemcmaster:
Another idea: when calling any project, we override CustomBeforeMicrosoftCommonTargets and CustomBeforeMicrosoftCommonCrossTargetingTargets to the location of the path containing dotnet-watch's targets.

hartmannr76:
[...]

PR #347, opened 2017-09-22
Stop using MSBuildProjectExtensionsPath in dotnet-watch and user-secrets

natemcmaster:
Resolves https://github.com/aspnet/DotNetTools/issues/244

Use CustomAfterMicrosoftCommonTargets instead of MSBuildProjectExtensionsPath.  

- No more need to write to obj/$(Project).g.dotnetwatch.targets  
- Works on project that have changed default file locations via BaseIntermediateOutputPath  

Simplify DotNetWatch targets  
- Condense to one targets file  
- Simplify dependency chain of targets  
- Build project references in a parallel


src/Microsoft.DotNet.Watcher.Tools/Internal/MsBuildFileSetFactory.cs
```diff
@@ -162,29 +143,31 @@ public class MsBuildFileSetFactory : IFileSetFactory
...
+            var args = new List<string>
             {
-                // ensure obj folder is available
-                Directory.CreateDirectory(Path.GetDirectoryName(projectExtensionFile));
+                "/nologo",
+                "/v:n",
+                "/t:" + TargetName,
+                "/p:DotNetWatchBuild=true", // extensibility point for users
+                "/p:DesignTimeBuild=true", // don't do expensive things
+                "/p:CustomAfterMicrosoftCommonTargets=" + watchTargetsFile,
```
bricelam: You can't specify multiple targets to include in this property, can you? This prevents projects from using this extensibility point.

natemcmaster: Yes, this would override the project's setting. From what I've seen, this property is rarely used, much less than BaseIntermediateOutputPath. Even when used, it seems unlikely to change the list of files in Compile/EmbeddedResource/ProjectReference.

Another option is to write the targets file to [MSBuildUserExtensionsPath](https://github.com/Microsoft/msbuild/blob/1db4e2b95b2336d5b7083341ee0bff9cf0e3c533/src/Tasks/Microsoft.Common.CurrentVersion.targets#L28), though, I have reservations about adding files to machine state, and this wouldn't work across major version changes in MSBuild.

bricelam: Sounds good. I just wanted to make sure you considered the tradeoffs.

natemcmaster: Btw, before I merge this, are you aware of any other extensibility points I should consider? These are the ones I've looked at:
 - MSBuildProjectExtensionsPath 
 - MSBuildUserExtensionsPath
 - CustomAfterMicrosoftCommonTargets
 - CustomBeforeMicrosoftCommonTargets
 - $(MSBuildProjectFullPath).user

bricelam: `Directory.Build.targets` :trollface:

natemcmaster: haha. Going with CustomAfterMicrosoftCommonTargets for now.

PR #355, opened 2017-11-07
Create an initial implementation of the secret manager for VS

natemcmaster:
cref https://github.com/aspnet/Coherence-Signed/issues/668

cc @AArnott


natemcmaster:
See also https://github.com/aspnet/Configuration/pull/750

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                store.Save();
+            }
+
+            return result;
+        }
+
+        private void EnsureKeyNameIsValue(string name)
```
AArnott: `EnsureKeyNameIsValue` -> `EnsureKeyNameIsValid`?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (string.IsNullOrEmpty(name))
```
AArnott: Why are you checking for null here again when you already tested that condition above?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var userSecretsId = await _projectProperties.GetEvaluatedPropertyValueAsync(UserSecretsPropertyName);
+
+            if (string.IsNullOrEmpty(userSecretsId))
+            {
+                userSecretsId = Guid.NewGuid().ToString();
+                await _projectProperties.SetPropertyValueAsync(UserSecretsPropertyName, userSecretsId);
+            }
+
+            return new SecretsStore(userSecretsId);
```
AArnott: Can you add `await TaskScheduler.Default;` here so that we know the disk I/O won't be on the UI thread?
Or, if you can't reference MS.VS.Threading.dll, instead of awaiting the task scheduler, you can use this syntax:

```csharp
return await Task.Run(() => new SecretsStore(userSecretsId));
```

Although IMO constructors should be fairly lightweight, so I would personally refactor this to have a `LoadAsync` static factory method on the `SecretsStore`, or keep all the load/save functionality in this outer class and just use a straight-up Dictionary.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            EnsureKeyNameIsValue(name);
+
+            var store = await GetOrCreateStoreAsync();
+
+            var result = store.Remove(name);
+
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                store.Save();
```
AArnott: This will overwrite changes other (logical) threads may also be doing. For example, this:

```csharp
await (Task.WhenAll(store.AddAsync(1,2), store.AddAsync(3,4)));
```

Would very likely end up writing a file with only one of the secrets in it instead of both. We'd prefer the API to be more robust than that, and the interface documentation indicates the implementation is expected to be thread-safe.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
AArnott: Similar change for saving the file.

AArnott: Given users can open the secrets.json file in the VS editor, and in fact a command is exposed in VS to do just that, I think it appropriate to follow the best practice of first trying to open the file from the running document table if it is present there, and then updating that buffer with the change if it was opened from there -- only falling back on I/O if the document isn't already open. That gives the *integrated* development environment feeling instead of running an operation in the IDE that then pops a dialog saying "you made a change outside VS--would you like to reload the file?" to the user.

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.Extensions.SecretManager.VisualStudio
+{
+    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [InstalledProductRegistration("#110", "#112", "1.0", IconResourceID = 400)]
+    [Guid(SecretManagerVSPackage.PackageGuidString)]
+    public sealed class SecretManagerVSPackage : Package
+    {
+        public const string PackageGuidString = "d6b82ee5-8f4b-41f2-a581-fefa7f57a239";
```
AArnott: Why do we need a VS package for this? That's not required for a MEF assembly. Is there some reason that's not obvious to me that we need this?

natemcmaster: Just following what File->New gave me and all other VSIX projects appear to have. This is my first VSIX and don't really understand how all this bootstrapping stuff works. If it's not required, I'll strip it out.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: Hmm, never heard of running document table. I'll look into this.

natemcmaster: Trying to add `await TaskScheduler.Default`. Compielr fails with "Cannot find the interop type that matches the embedded interop type 'Microsoft.VisualStudio.Shell.Interop.IVsTask'. Are you missing an assembly reference?"

Microsoft.VisualStudio.Threading is referenced. Any idea what I'm missing?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: This will be problematic. I'll have to think about this more so we can avoid race conditions on reading from/writing to disk

shared/SecretManager/SecretsStore.cs
```diff
@@ -0,0 +1,97 @@
...
+            File.WriteAllText(secretsFilePath, contents.ToString(), Encoding.UTF8);
+        }
+
+        protected virtual Dictionary<string, string> Load(string userSecretsId)
+        {
+            var secretsFilePath = PathHelper.GetSecretsPathFromSecretsId(userSecretsId);
+
+            // workaround bug in configuration that causes the JSON loader to fail
+            var secretDir = Path.GetDirectoryName(secretsFilePath);
+            Directory.CreateDirectory(secretDir);
```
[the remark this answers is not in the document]
natemcmaster: It was a bug in the JSON loader. I think this workaround can be removed now as that bug has been fixed.

shared/SecretManager/SecretsStore.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public virtual bool Remove(string key)
+        {
+            if (_secrets.ContainsKey(key))
+            {
+                _secrets.Remove(key);
+                return true;
+            }
+
+            return false;
```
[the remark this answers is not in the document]
natemcmaster: .Remove only edits the in-memory collection. It doesn't persist the change to disk until you call .Save()

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
natemcmaster: Seems it's required. Without it, build fails:

Error		CreatePkgDef : error : ArgumentException: No Visual Studio registration attribute found in this assembly.
The assembly should contain an instance of the attribute 'Microsoft.VisualStudio.Shell.RegistrationAttribute' defined in assembly 'Microsoft.VisualStudio.Shell.Framework' version '15.0.0.0' 
   at Microsoft.VisualStudio.Tools.CreatePkgDef.ProcessAssembly(String fileName, Hive hive, PkgDefContext context, Boolean register, RegistrationMode mode)
   at Microsoft.VisualStudio.Tools.CreatePkgDef.DoCreatePkgDef(InputArguments inputArguments)
   at Microsoft.VisualStudio.Tools.CreatePkgDef.Main(String[] arguments)	SecretManager.VisualStudio

build/VSIX.targets
```diff
@@ -0,0 +1,86 @@
+﻿<Project>
```
rynowak: Should we commonize these things? We have similar MSBuild in Razor.

shared/SecretManager/SecretsStore.cs
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Microsoft.Extensions.SecretManager
+{
+    /// <summary>
+    /// Provides read and write access to the secrets.json file for local user secrets.
+    /// This is not thread-safe.
+    /// This object is meant to have a short lifetime.
+    /// When calling <see cref="Save"/>, this will overwrite the secrets.json file. It does not check for concurrency issues if another process has edited this file.
+    /// </summary>
+    internal class SecretsStore : IReadOnlyDictionary<string, string>
```
rynowak: ugh - plurals in names always make a mess of things. Can this just be `SecretStore`?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
rynowak: Sounds like you have a using for `Microsoft.VisualStudio.Shell`. There's a common name conflict there.

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
rynowak: I can help you try to figure this out tomorrow

shared/SecretManager/SecretsStore.cs
```diff
@@ -0,0 +1,90 @@
...
+                foreach (var secret in _secrets.AsEnumerable())
+                {
+                    contents[secret.Key] = secret.Value;
+                }
+            }
+
+            File.WriteAllText(secretsFilePath, contents.ToString(), Encoding.UTF8);
+        }
+
+        protected virtual Dictionary<string, string> Load(string userSecretsId)
```
rynowak: It might be worth factoring all of the save/load stuff into it's own internal interface.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Extensions.SecretManager.VisualStudio
```
rynowak: Not sure if naming was dicussed yet, but normally something that provides visual studio integration would be named like `Microsoft.VisualStudio...`.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+{
+    public class ProjectLocalSecretsManager : IVsProjectSecrets, SVsProjectLocalSecrets
```
rynowak: Does this class *need* to be public?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        private readonly IProjectPropertiesProvider _propertiesProvider;
+
+        public ProjectLocalSecretsManager(IProjectPropertiesProvider propertiesProvider)
+        {
+            _propertiesProvider = propertiesProvider ?? throw new ArgumentNullException(nameof(propertiesProvider));
+        }
+
+        public string SanitizeName(string name) => name;
+
+        public IReadOnlyCollection<char> GetInvalidCharactersFrom(string name) => Array.Empty<char>();
```
rynowak: Is this right? I see that there's no sanitization or filtering, but then line 32 checks the validity 😕

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+            EnsureKeyNameIsValid(name);
+            var store = await GetStoreAsync();
+            if (store.ContainsKey(name))
+            {
+                throw new ArgumentException("A secret with this name already exists.", nameof(name));
+            }
+
+            store.Set(name, value);
+            cancellationToken.ThrowIfCancellationRequested();
+            store.Save();
```
rynowak: This class would be the place to introduce some synchronization

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public async Task<string> GetSecretAsync(string name, CancellationToken cancellationToken = default)
+        {
+            EnsureKeyNameIsValid(name);
+            var store = await GetStoreAsync();
+            return store[name];
+        }
+
+        public async Task<IReadOnlyCollection<string>> GetSecretNamesAsync(CancellationToken cancellationToken = default)
```
rynowak: @AArnott - is there any expectation or provision for something like transactions on secret? 

For instance if a dialog is showing in VS that's been populated using one of the APIs like `GetSecretNamesAsync` and the file is updated using the commandline tools, what should happen? 

Are we assuming that those kinds of things are out of scope or not important?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (string.IsNullOrEmpty(userSecretsId))
+            {
+                // TODO how should this be handled?
+                throw new InvalidOperationException("UserSecrets cannot be read or altered on this project because 'UserSecretsId' is not set");
+            }
+
+            // TODO figure out why this API causes compiler error:
+            // "Cannot find the interop type that matches the embedded interop type 'Microsoft.VisualStudio.Shell.Interop.IVsTask'."
+             //await TaskScheduler.Default;
```
rynowak: This is a conflict between `Microsoft.VisualStudio.Shell.Task` and `System.Threading.Tasks.Task`. I make it a point to just alias everything I use from the VS shell namespace if I'm mixing shell and async code.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
rynowak: You mentioned to me some concerns about interacting with open files and the RDT. We can talk about this and figure it out tomorrow.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        private async Task<SecretsStore> GetStoreAsync()
+        {
+            var userSecretsId = await _propertiesProvider.GetCommonProperties().GetEvaluatedPropertyValueAsync(UserSecretsPropertyName);
+
+            if (string.IsNullOrEmpty(userSecretsId))
+            {
+                // TODO how should this be handled?
+                throw new InvalidOperationException("UserSecrets cannot be read or altered on this project because 'UserSecretsId' is not set");
```
rynowak: I think the best thing to do would be to be as tolerant as possible.

If the project doesn't have a secrets ID and it's a 'read' then obviously the answer is 'emtpy. If the project doesn't have a secrets id and it's a 'write' we should try to generate one.

tooling/SecretManager.VisualStudio/SecretManager.VisualStudio.csproj
```diff
@@ -0,0 +1,136 @@
...
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.Configuration.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.Configuration.FileExtensions.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.Configuration.Json.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.Configuration.UserSecrets.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.FileProviders.Abstractions.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.FileProviders.Physical.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.FileSystemGlobbing.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Microsoft.Extensions.Primitives.dll" />
+    <VSIXSourceItem Include="$(OutputPath)Newtonsoft.Json.dll" />
+    <Content Include="@(VSIXSourceItem)">
```
rynowak: Yeah isn't this a super **FUN** workaround?

tooling/SecretManager.VisualStudio/SecretManagerFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Extensions.SecretManager.VisualStudio
+{
+    internal class SecretManagerFactory 
+    {
+        private readonly Lazy<ProjectLocalSecretsManager> _secretManager;
+        private readonly ConfiguredProject _project;
+
+        [ImportingConstructor]
+        public SecretManagerFactory(ConfiguredProject project)
```
rynowak: Is configured project the right scope for this? I would expect it to be unconfigured project.

ConfiguredProject = 1 instance per project file X debug or release X bitness
UnconfiguredProject = 1 instance per projectfile

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
rynowak: I'll take a look at this tomorrow with you and see if I can figure it out. @AArnott do you have an idea if there's something obvious we need to do besides delete the VSPackage and remove the entry from the vsixmanifest?

tooling/SecretManager.VisualStudio/SecretManagerFactory.cs
natemcmaster: Yes. Although this is uncommon, the following would be a legitimate way to use UserSecrets.
```xml
<PropertyGroup>
  <UserSecretsId Condition="'$(Configuration)' == 'Debug'">debug-secrets-id</UserSecretsId>
  <UserSecretsId Condition="'$(Configuration)' == 'Release'">release-secrets-id</UserSecretsId>
</PropertyGroup>
```

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: I decided to added a throw here for now. Initially I was generating a guid and calling IProjectProperties.SetPropertyValueAsync. I pulled it out for now to avoid race conditions. Also, we need to address how this surfaces to the user in the UI.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: I have this at the top of the file: `using Task = System.Threading.Tasks.Task;`. Is that the issue?

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: Yes, this is right. EnsureKeyNameIsValid is just asserting string.IsNullOrEmpty. secrets.json allows any character as the config key.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: no

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: hasn't been discussed. What about `Microsoft.VisualStudio.Extensions.SecretManager`?

rynowak: I would drop `Extensions` since it means something else in the VS context.

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
rynowak: I'd try nuking the `using Microsoft.VisualStudio.Shell` and see if that fixes this problem (while causing others)

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
rynowak: > Also, we need to address how this surfaces to the user in the UI.

It depends on the gesture but generally you would not surface this in the UI. If tooling can trivially fix a problem, it should.

tooling/SecretManager.VisualStudio/SecretManagerFactory.cs
rynowak: Only if we decide to support it.

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
AArnott: There is a project property to disable pkgdef generation. 
However, you *do* want a pkgdef, since there is one VS registration attribute your assembly should always have:

```csharp
[assembly: ProvideCodeBase(CodeBase = @"$PackageFolder$\YOUR.ASSEMBLY.NAME.dll")]
```

That way, VS can find your assembly for sure instead of hoping. It improves overall reliability and even performance in loading your assembly. That should eliminate the pkgdef build error since you'll now have an attribute to create a pkgdef for. :)

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
AArnott: Well, we do expect the returned collection to be frozen -- not a live one that keeps changing. That is somewhat a given since the interface doc says it must be thread safe, and a collection that changes is not thread-safe to return.

But the idea of knowing when to refresh is an interesting question I hadn't thought about when I designed the interface. I think we should just table it for now though, since it would undoubtedly be costly for you to implement and we don't yet have a scenario for it -- we might eventually, but we can add another interface when and if the time comes.

tooling/SecretManager.VisualStudio/SecretManagerVSPackage.cs
natemcmaster: Awesome, thanks for the pointer. That worked.

build/VSIX.targets
natemcmaster: maybe one day. For now, we only have this in two places. If using "full" MSBuild becomes a more common requirement for building, we may look into having a switch to run korebuild on full MSBuild from end to end instead

natemcmaster:
:up: :date:

Added thread-safety, and I took a whack at IVsInvisibleEditor/RDT for file IO.

@rynowak is currently working on a dummy UI project so we can actually test and validate this.

AArnott:
Glad to hear of the progress. I'll re-review on Monday.

rynowak:
> Glad to hear of the progress. I'll re-review on Monday.

Keep in mind we haven't gotten to 'hello world' here yet, so you may find obvious mistakes 😆

tooling/SecretManager.VisualStudio/Sources/JsonConfigurationFileParser.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    internal class JsonConfigurationFileParser
+    {
+        private readonly IDictionary<string, string> _data = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        private readonly Stack<string> _context = new Stack<string>();
+        private string _currentPath;
+
+        private JsonTextReader _reader;
+
+        public IDictionary<string, string> Parse(Stream input)
```
rynowak: This method should be static

natemcmaster:
:up: :date: Ok, I've verified with our test GUI that can read/write to the secrets.json file.

Note - RDT and IVsInvisibleEditor made SecretSecret store more complicated than I think it should be for simple read/write operations. I ran into issues that I don't know how to fix, exceptions with error messages that didn't contain useful info, and require APIs I couldn't find any docs for. System.IO.File appears to work just fine for now. If a user had an editor open to the secrets.json file, they will get the "Reload file" dialog when the file changes on disk.

tooling/SecretManager.VisualStudio/Sources/JsonConfigurationFileParser.cs
natemcmaster: Can't be. The parser has context that is unique per instance.

rynowak: Yeah, this method should new up the parser 👍

natemcmaster: It's worth pointing out this is an exact copy of https://github.com/aspnet/Configuration/blob/dev/src/Config.Json/JsonConfigurationProvider.cs.

I'm fine with the change, but want to avoid making changes here unless we plan to port back to aspnet/Configuration. Can we do that change in aspnet/Configuration first and then port it here?

DotNetTools.sln
```diff
@@ -44,42 +44,90 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Watcher.To
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Caching.SqlConfig.Tools", "src\Microsoft.Extensions.Caching.SqlConfig.Tools\Microsoft.Extensions.Caching.SqlConfig.Tools.csproj", "{53F3B53D-303A-4DAA-9C38-4F55195FA5B9}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.DeveloperCertificates.Tools", "src\Microsoft.AspNetCore.DeveloperCertificates.Tools\Microsoft.AspNetCore.DeveloperCertificates.Tools.csproj", "{4FED5119-EE5C-4753-88A4-D61BDEB4D6C8}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.DeveloperCertificates.Tools", "src\Microsoft.AspNetCore.DeveloperCertificates.Tools\Microsoft.AspNetCore.DeveloperCertificates.Tools.csproj", "{4FED5119-EE5C-4753-88A4-D61BDEB4D6C8}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SecretManager.VisualStudio", "tooling\SecretManager.VisualStudio\SecretManager.VisualStudio.csproj", "{5E117F2E-7152-447F-BF47-59F759EEF3A7}"
```
rynowak: Any objection to changing this project to match the assembly name? `Microsoft.VisualStudio.SecretManager` IIRC

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            EnsureKeyNameIsValid(name);
+            await TaskScheduler.Default;
+
+            using (await _semaphore.EnterAsync(cancellationToken))
+            using (var store = await GetOrCreateStoreAsync(cancellationToken))
+            {
+                if (store.ContainsKey(name))
+                {
+                    throw new ArgumentException("A secret with this name already exists.", nameof(name));
```
rynowak: This should use the resources file

tooling/SecretManager.VisualStudio/SecretManagerFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            _secretManager = new Lazy<ProjectLocalSecretsManager>(() =>
+            {
+                var propertiesProvider = _project.Services.ActiveConfiguredProjectProvider.ActiveConfiguredProject.Services.ProjectPropertiesProvider;
+                return new ProjectLocalSecretsManager(propertiesProvider, serviceProvider);
+            });
+        }
+        
+        [ExportVsProfferedProjectService(typeof(SVsProjectLocalSecrets))]
+        [AppliesTo(ProjectCapabilities.AlwaysApplicable)]
```
rynowak: This needs to use our capability. I think I may have changed this while I was hacking things 😆

natemcmaster: Good catch. Updated

tooling/SecretManager.VisualStudio/ProjectLocalSecretsManager.cs
natemcmaster: Done

DotNetTools.sln
natemcmaster: Nope. Done.

tooling/SecretManager.VisualStudio/Sources/JsonConfigurationFileParser.cs
natemcmaster: Updated to be static. I'll add this change to https://github.com/aspnet/Configuration/pull/750 as well

natemcmaster:
:up: :date:

AArnott:
@natemcmaster I don't see any follow-up changes for RDT integration that I asked for. Is that going to be done for 15.7?
Also, as this serializer writes the file 'flattened' (no nesting structure for connection strings, etc.) will the ASP.NET APIs that convey this information as "connection strings" and other structures. @csharpfritz has more info on this if you need.

natemcmaster:
> I don't see any follow-up changes for RDT integration that I asked for. Is that going to be done for 15.7?

No plans at the moment. I ran out of time and found the RDT api too daunting. I spent a few days on it but failed to get an RDT implementation that was working well in all scenarios. We figured let's start with the simple thing that works first and then we can look into RDT if we run into issues.

> Also, as this serializer writes the file 'flattened' (no nesting structure for connection strings, etc.) will the ASP.NET APIs that convey this information as "connection strings" and other structures. 

Yes, ASP.NET Core can parse structure even though the JSON file is flattened.

AArnott:
Thanks. In a meeting today, we discussed the possibility of lacking RDT integration for the first release. Considering we imagine a UX where the user isn't likely to have to interact with the file directly eventually, investing in the RDT in the short-term would have short-lived gains. And the side-effect of the "reload doc?" popup in the short term doesn't seem _too_ bad for a short-lived problem. So I think I'm OK with leaving RDT out.

CC: @csharpfritz @danroth27 @glennc (and Steve Molloy who I don't have a github username for)

natemcmaster:
Yeah, that matches with our thinking, too. We are hoping the UI will reduce the need to edit the user-secrets JSON file directly in VS, so RDT may not produce a whole lot of value right away. The fact that user-secrets are stored as a JSON file on the machine is meant to be an implementation detail, anyways.

PR #356, opened 2017-11-10
Add support for detecting when the user cancels the trust prompt and a mode to check for the validity and trust of existing certificates

javiercn:


src/Microsoft.AspNetCore.DeveloperCertificates.Tools/Program.cs
```diff
@@ -44,6 +45,11 @@ public static int Main(string[] args)
                         "Password to use when exporting the certificate with the private key into a pfx file",
                         CommandOptionType.SingleValue);
 
+                    var check = c.Option(
+                        "-c|--check",
```
natemcmaster: Seems like this should be its own verb. The usage is starting to look messy. What about splitting the behavior into sub commands?
```
dotnet developercertificates list
dotnet developercertificates generate
```

javiercn: I wanted to do it like that in the beginning, but when we have the meeting to discuss this the PMs wanted something simpler (1 uber-command vs a 'crud' api). So that's why it is this way. We can change it later if we get negative feedback on this.

natemcmaster: Which PM can I bug about this? This is bad design.

javiercn: Dan and Glenn.

I don't think it's that bad. An analogous example is PowerShell commands that offer a --whatif parameter. In this case --check == --whatif

natemcmaster: `--what-if` and `--dry-run` commands typically execute the same code path, just skipping the "commit" action. But in this case `--check` is entering a whole new code path, so doesn't really fit the criteria IMO. In this current proposal, `--check` is a combination of `generate --whatif` and `list`.

javiercn: @danroth27 came to talk to me. His decision is to keep it as is for right now. I'll file a bug to track the discussion.

natemcmaster: Sigh. This is how good UX dies....on backlogged issue at a time.

natemcmaster: My beef with this approach in general is that we're deferring issue until later. I know I'm being more salty than usual, but it's hard to change stuff once it's in production. I really think `developercertificates` is a bad verb. It's too long. And `--check` is ambiguous. 

A "meta" verb is confusing and hard to use. Splitting into subcommands makes it much easier to navigate the various actinos. 

Look at `sn.exe` for a bad example. It has half a dozen functions. You have to workout the right combinations of flags by reading the help carefully. I know we're far off from the amount of options in `sn`, but this is the road you're on.

```
Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Usage: sn [-q|-quiet] <option> [<parameters>]
 Options:
  -a identityPublicKeyFile identityKeyPairFile signaturePublicKeyFile
    Generate AssemblySignatureKeyAttribute data to migrate the identity key to the signature key from a file.
  -ac identityPublicKeyFile identityKeyPairContainer signaturePublicKeyFile
    Generate AssemblySignatureKeyAttribute data to migrate the identity key to the signature key from a key container.
  -c [<csp> [<provtype>] ]
    Set/reset the name of the CSP to use for MSCORSN operations.
  -d <container>
    Delete key container named <container>.
  -dg <assembly> <digest file> [-rehash]
    Compute the digest of an assembly, generating a digest file for signing by
    the -ds or -dsc commands.
    If -rehash is specified, modules are re-hashed before digest calulation
  -ds <digest file> <keyfile> [-ecma]
    Sign a digest file generated by the -dg command using the full key pair
    from the given key file.
    If -ecma is used, keyfile is treated as the real key for ECMA signing.
  -dsc <digest file> <container> [-ecma]
    Sign a digest filegenerated by the -dg command using the key in the
    specified key container.
    If -ecma is used, container is treated as the real key for ECMA signing.
  -di <digest file> <assembly>
    Insert the signature from the digest file signed by the -ds or -dsc
    commands into the assembly.
  -dh <digest file> <hash file>
    Extract the base64 encoded hash value from the digest file
  -du <signature file> <digest file>
    Update the signature in the digest file with a base64 encoded signature
    in a signature file. WARNING: using this operation may result in a digest
    file that contains an invalid signature for its corresponding assembly.
  -dd <digest file>
    Dump the digest file to the console
  -D <assembly1> <assembly2>
    Verify <assembly1> and <assembly2> differ only by signature.
  -e <assembly> <outfile>
    Extract public key from <assembly> into <outfile>.
  -i <infile> <container>
    Install key pair from <infile> into a key container named <container>.
  -k [<keysize>] <outfile>
    Generate a new key pair of the specified size and write it into <outfile>.
  -m [y|n]
    Enable (y), disable (n) or check (no parameter) whether key containers
    are machine specific (rather than user specific).
  -o <infile> [<outfile>]
    Convert public key in <infile> to text file <outfile> with comma separated
    list of decimal byte values.
    If <outfile> is omitted, text is copied to clipboard instead.
  -p <infile> <outfile> [<hashalg>]
    Extract public key from key pair in <infile> and export to <outfile>,
    embedding the specified hash algorithm (sha1|sha256|sha384|sha512).
  -pc <container> <outfile> [<hashalg>]
    Extract public key from key pair in <container> and export to <outfile>,
    embedding the specified hash algorithm (sha1|sha256|sha384|sha512).
  -Pb [y|n]
    Enable (y), disable (n) or check (no parameters) the CLR policy allowing
    trusted applications to bypass strong name signature verification on their
    assemblies.
  -q
    Quiet mode. This option must be first on the command line and will suppress
    any output other than error messages.
  -R[a] <assembly> <infile> [-ecma]
    Re-sign signed or partially signed assembly with the key pair in <infile>.
    If -Ra is used, hashes are recomputed for all files in the assembly.
    If -ecma is used, infile is treated as the real key for ECMA signing.
  -Rc[a] <assembly> <container> [-ecma]
    Re-sign signed or partially signed assembly with the key pair in the key
    container named <container>.
    If -Rca is used, hashes are recomputed for all files in the assembly.
    If -ecma is used, container is treated as the real key for ECMA signing.
  -Rh <assembly>
    Re-compute hashes for all files in the assembly.
  -t[p] <infile>
    Display token for public key in <infile> (together with the public key
    itself if -tp is used).
  -T[p] <assembly>
    Display token for public key of <assembly> (together with the public key
    itself if -Tp is used).
  -TS <assembly> <infile>
    Test-sign signed or partially signed assembly with the key pair in
    <infile>.
  -TSc <assembly> <container>
    Test-sign signed or partially signed assembly with the key pair in the key
    container named <container>.
  -v[f] <assembly> [{-ecmakey <keyfile> | -ecmacontainer <container>}]
    Verify <assembly> for strong name signature self consistency. If -vf is
    specified, force verification even if disabled in the registry.
    If -ecmakey is specified, keyfile is treated as the real ECMA key.
    If -ecmacontainer is specified, container is treated as the real ECMA key.
  -Vk <regfile> <assembly> [<userlist>] [<testkey>]
    Generate a registry script in <regfile> to register <assembly> for
    verification skipping (with an optional, comma separated list of usernames
    for which this will take effect and an optional test public key in
    <testkey>). <assembly> can be specified as * to indicate all assemblies or
    *,<public key token> to indicate that all assemblies with the given public
    key token. Public key tokens should be specified as a string of hex digits.
  -Vl
    List current settings for strong name verification on this machine.
  -Vr <assembly> [<userlist>] [<infile>]
    Register <assembly> for verification skipping (with an optional, comma
    separated list of usernames for which this will take effect and an
 optional test public key in <infile>).
 <assembly>    can be specified as * to indicate all assemblies or *,<public key token> to
    indicate that all assemblies with the given public key token. Public key
    tokens should be specified as a string of hex digits.
  -Vu <assembly>
    Unregister <assembly> for verification skipping. The same rules for
    <assembly> naming are followed as for -Vr.
  -Vx
    Remove all verification skipping entries.
  -?
  -h
    Displays this help text.
```

danroth27: The user requirement here is to have one simple command that the user can copy and past from an error message in the rare event that they don't already have the HTTP dev cert. As I understand it this `check` parameter is specifically for VS tooling scenarios and isn't really something we expect users to use, so I'm not really fussed about it.

javiercn: @natemcmaster I understand your feedback and its captured here https://github.com/aspnet/DotNetTools/issues/357.

Right now the priority is to make the whole E2E scenario work. We will be reviewing this in the following weeks and I'll make sure I bring this up, but for now I would like to stick with the decision that was made at the time. In the end, its up to the PMs to decide that experience they want to have.

natemcmaster:
let me pivot from ranting on CLI verbs to build stuff. It looks like compilation is failing. The new .sources packages should be available now. Once you've update the package versions so this repo passes on AppVeyor/Travis, go ahead and :shipit:

PR #361, opened 2017-11-16
Update samples and tests to target netcoreapp2.1

pranavkm:


pranavkm:
@natemcmaster can you take over this PR? I'm not super familiar with the tests here and it's probably easier for you to debug this.

natemcmaster:
For this repo, I don't think it makes sense to multitarget the tests. The console apps in this project will only ever be executed against netcoreapp2.0. We should keep the tests and the app TFM aligned.

pranavkm:
> The console apps in this project will only ever be executed against netcoreapp2.0. 

I don't follow. Is this the actual tools?

natemcmaster:
dotnet-watch is a netcoreapp2.0 console app. There is no reason to run it as a netcoreapp2.1 and netcoreapp2.0 in tests.

https://github.com/aspnet/DotNetTools/blob/382cfa8e074bfbef884ea5f5a7e1add5e661258a/src/Microsoft.DotNet.Watcher.Tools/Microsoft.DotNet.Watcher.Tools.csproj#L4-L5

pranavkm:
> dotnet-watch is a netcoreapp2.0 console app.

Sure, and it stays that way. But the premise is to test it against apps running on different runtimes.

natemcmaster:
But why test against a runtime that the app never uses? This isn't a class library: it's an app we deploy via DotNetCliToolReference (for now.)

pranavkm:
I'll let you take over the changes for this repo. I'm not entirely convinced why we don't need to verify that the tool works against an app running on netcoreapp2.1, but you know the repo better.

natemcmaster:
:up: :date:

natemcmaster:
I verified that a package targeting netcoreapp2.1 can install and run as a DotNetCliToolReference. I've upgraded everything in the repo to use netcoreapp2.1 now.

pranavkm:
:shipit:

natemcmaster:
Hmm, tests are failing on macOS and Linux. I'll investigate and try to resolve before merging.

natemcmaster:
Windows and macOS tests pass now, but tests still fail on Travis. I can't reproduce failures locally, but they are consistent on Travis.

natemcmaster:
Tests pass! Merging to dev now

PR #372, opened 2017-12-20
Mark the SecretManager VSIX as IsProductComponent=true

natemcmaster:
Generates the json manifest required for signing and VS insertion.

### DIFF
--- a/tooling/Microsoft.VisualStudio.SecretManager/Microsoft.VisualStudio.SecretManager.csproj
+++ b/tooling/Microsoft.VisualStudio.SecretManager/Microsoft.VisualStudio.SecretManager.csproj
@@ -3,6 +3,12 @@
   <PropertyGroup>
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+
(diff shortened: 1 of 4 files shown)


tooling/Microsoft.VisualStudio.SecretManager/Microsoft.VisualStudio.SecretManager.csproj
```diff
@@ -3,6 +3,12 @@
   <PropertyGroup>
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+
+    <!-- Required to place the vsix in the right location by WTE MSI -->
+    <ExtensionInstallationRoot>CommonExtensions</ExtensionInstallationRoot>
+    <ExtensionInstallationFolder>Microsoft\AspNetCore</ExtensionInstallationFolder>
```
natemcmaster: @joeloff totally just guessed on this one. What's the right path?

joeloff: That should be fine. The only real catch is picking a too generic name, e.g. if we ever have another AspNetCore related extension, then we'd have a problem or have to come up with a different location for the new extension

natemcmaster: Updated

PR #377, opened 2018-01-09
Fix FunctionalTests on fresh system

ryanbrandenburg:
Some tests failed on systems that didn't have packages already installed because their NuGet.config was empty (totally fresh ones or those that delete their .dotnet folder). I tested this using docker.


test/dotnet-watch.FunctionalTests/TestProjects/NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
natemcmaster: .NET Core packages may not be on myget.org when our tests run (e.g. in an orchestrated build). We should generate this file instead based on the `RestoreSources` property used at build-time. We've done this in other places,... for example, in aspnet/Scaffolding.

https://github.com/aspnet/Scaffolding/blob/dev/test/Shared/MsBuildProjectStrings.cs.in#L16
https://github.com/aspnet/Scaffolding/blob/dev/test/Directory.Build.targets#L23-L30

ryanbrandenburg:
:up::date:

Issue #383, opened 2018-01-17
Update to global tools latest nupkg format


wli3:
Need to update to global tools latest nupkg format. Or the result nupkg cannot be consumed by CLI once https://github.com/dotnet/cli/pull/8414 is merged.  I am referencing [dotnet-watch.nuspec](https://github.com/aspnet/DotNetTools/blob/dev/src/dotnet-watch/dotnet-watch.nuspec) 

- [ ]  [DotnetToolSettings.xml](https://github.com/aspnet/DotNetTools/blob/dev/src/dotnet-watch/dotnet-watch.nuspec#L24) need to store under TFM/any. [Reference](https://github.com/wli3/cli/blob/9c2a9561850e3cbf3b7db11de1cdc7d44c2745c5/test/Microsoft.DotNet.ToolPackage.Tests/SampleGlobalTool/includepublish.nuspec#L17)

- [ ]  Need to add _Microsoft.NETCore.Platforms_ as dependency. The version of the package should be the runtime version/offline package version of Microsoft.NETCore.Platforms. So CLI can find that package from offline cache. [Reference](https://github.com/wli3/cli/blob/9c2a9561850e3cbf3b7db11de1cdc7d44c2745c5/test/Microsoft.DotNet.ToolPackage.Tests/SampleGlobalTool/includepublish.nuspec#L12) (Note version should **not** be 2.0.1. It won’t be in offline cache)

natemcmaster:
>Need to add Microsoft.NETCore.Platforms as dependency.

I don't understand this requirement. Can you clarify? Is this something all global tools packages will have to carry? Is this a workaround?

natemcmaster:
cc @muratg

wli3:
RID is transparent to NuGet, to select correct RID, `Microsoft.NETCore.Platforms` is required to select right RID. In portable case, this package is still needed to ANY given consumer's RID say win-x64. Detail is [here](https://github.com/dotnet/designs-microsoft/blob/dcd81fd56d0b7bcb1bbca3872614bef2778d6f56/proposals/implementation_global_tools.md#handling-self-contained-applications).

Pack tool experience will add this dependency implicitly so producer using pack experience do not need to worry about this.

natemcmaster:
I don't think this is something we want to put into our tool nuspec. It's not actually a dependency of the tool. It seems to me like having Microsoft.NETCore.Platforms in the restore graph is just an implementation detail of how `dotnet install tool` is implemented.

[...]

PR #385, opened 2018-01-24
React to changes in package layout requirements for global CLI tools

natemcmaster:
React to  dotnet/cli#8414 which causes latest dotnet-watch packages not to install via dotnet-install-tool.

Depends on https://github.com/aspnet/BuildTools/pull/558

Resolves https://github.com/aspnet/DotNetTools/issues/383


wli3:
any/any....

PR #392, opened 2018-02-08
Include symbols in build output for the VSIX project, and update vsix manifest dependency version

natemcmaster:
cc @NTaylorMullen @rynowak  - we should make the same change for Microsoft.VisualStudio.RazorExtension.pdb. Apparently, its not supposed to be in the VSIX itself.


rynowak:
@NTaylorMullen can you make the equivalent change for razor please?

tooling/Microsoft.VisualStudio.SecretManager/Microsoft.VisualStudio.SecretManager.csproj
```diff
@@ -37,7 +37,7 @@
     <IncludeDebugSymbolsInVSIXContainer>false</IncludeDebugSymbolsInVSIXContainer>
```
natemcmaster: @NTaylorMullen I think this is set to true in razor. Should be set to false IIUC

PR #401, opened 2018-02-28
Remove Microsoft.NETCore.Platforms from tool package nuspec

### LINKED ISSUE #400: Remove Microsoft.NETCore.Platforms package in nuspec of tools
No need to add `Microsoft.NETCore.Platforms` package to nspec anymore

https://github.com/dotnet/cli/issues/8617

Reference https://github.com/dotnet/cli/issues/8418

natemcmaster:
Resolves #400

### DIFF
--- a/src/PackGlobalTool.targets
+++ b/src/PackGlobalTool.targets
@@ -12,7 +12,7 @@
         description=$(Description);
         repositoryUrl=$(RepositoryUrl);
         targetframework=$(TargetFramework);
-        MicrosoftNETCorePlatformsPackageVersion=$(MicrosoftNETCorePlatformsPackageVersion);
+        repositoryCommit=$(RepositoryCommit);
       </NuspecProperties>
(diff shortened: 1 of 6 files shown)


src/PackGlobalTool.targets
```diff
@@ -12,7 +12,7 @@
         description=$(Description);
         repositoryUrl=$(RepositoryUrl);
         targetframework=$(TargetFramework);
-        MicrosoftNETCorePlatformsPackageVersion=$(MicrosoftNETCorePlatformsPackageVersion);
+        repositoryCommit=$(RepositoryCommit);
```
prafullbhosale: Is this a new thing that's needed?

natemcmaster: Yes, it's new feature of NuGet we started using last week. It'ss unrelated to #400, but figured I'd fix it as long as I'm editing the nuspec.

PR #411, opened 2018-03-21
React to changes in NuGetPackageVerifier

natemcmaster:
React to https://github.com/aspnet/BuildTools/pull/616.

Discovered as a by-product of upgrading to Newtonsoft.Json 11.

### DIFF
--- a/NuGetPackageVerifier.json
+++ b/NuGetPackageVerifier.json
@@ -27,6 +27,20 @@
             "tools/any/any/runtimes/unix/lib/netstandard2.0/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg.",
             "tools/any/any/runtimes/win/lib/netcoreapp2.0/System.Text.Encoding.CodePages.dll": "Assembly is built by another project but bundled in our nupkg.",
             "tools/any/any/runtimes/win/lib/netstandard2.0/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg."
+          },
+          "ASSEMBLY_FILE_VERSION_MISMATCH": {
+            "tools/any/any/System.Text.Encoding.CodePages.dll": "Assembly is built by another project but bundled in our nupkg.",
+            "tools/any/any/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg.",
+            "tools/any/any/runtimes/win/lib/netstandard2.0/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg.",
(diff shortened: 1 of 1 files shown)


NuGetPackageVerifier.json
```diff
@@ -27,6 +27,20 @@
             "tools/any/any/runtimes/unix/lib/netstandard2.0/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg.",
             "tools/any/any/runtimes/win/lib/netcoreapp2.0/System.Text.Encoding.CodePages.dll": "Assembly is built by another project but bundled in our nupkg.",
             "tools/any/any/runtimes/win/lib/netstandard2.0/System.Data.SqlClient.dll": "Assembly is built by another project but bundled in our nupkg."
+          },
+          "ASSEMBLY_FILE_VERSION_MISMATCH": {
+            "tools/any/any/System.Text.Encoding.CodePages.dll": "Assembly is built by another project but bundled in our nupkg.",
```
ryanbrandenburg: I wonder if we could add * exclusions to NPV. We don't ever care if there's an ASSEMBLY_FILE_VERSION_MISMATCH or ASSEMBLY_VERSION_MISMATCH on a System.* assembly right? That might help reduce the clutter in these files.

natemcmaster: NGPV supports "*" to completely exclude all warnings for that error, but we still want to check our own assemblies. I don't think it supports globbing patterns. Would be nice if we could specify "System.*"

natemcmaster: update to comment -- formatting

ryanbrandenburg: I was thinking even more agressive, we could include an exclusion for `System.*` in the actual NGPV tool since that's true of all our packages/repos.

natemcmaster: I think we're more likely to miss legitimate errors with something like that. I'd rather have a little clutter in a file we rarely update than miss legitimate warnings.

ryanbrandenburg: What would be an instance where we care about ASSEMBLY_FILE_VERSION_MISMATCH for a System.* package? Maybe I misunderstand our goals here. Because mostly when people see NPV warnings for System.* packages they just ignore them without much/any investigation. If the answer is "It's just not worth the dev time to make the change" then I understand, but if we actually care about System.* verification then I think it's likely we've made a lot of mistakes in that area already.

natemcmaster: If we ever wrote an assembly named System.\*. We've done that before - example the Microsoft.AspNet.WebApi.Client package.

ryanbrandenburg: Fair deuce.

PR #418, opened 2018-03-31
Disable node reuse from dotnet-watch and dotnet-user-secrets

natemcmaster:
At the moment, this is a guess. I suspect the flakiness may be due to the addition of node reuse to dotnet msbuild.

### DIFF
--- a/src/dotnet-user-secrets/Internal/ProjectIdResolver.cs
+++ b/src/dotnet-user-secrets/Internal/ProjectIdResolver.cs
@@ -55,7 +55,11 @@ namespace Microsoft.Extensions.SecretManager.Tools.Internal
                     FileName = DotNetMuxer.MuxerPathOrDefault(),
                     Arguments = ArgumentEscaper.EscapeAndConcatenate(args),
                     RedirectStandardOutput = true,
-                    RedirectStandardError = true
+                    RedirectStandardError = true,
+                    Environment =
+                    {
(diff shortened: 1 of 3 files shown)


Eilon:
If we do this, do we need to log a bug to undo this as well? Presumably this isn't meant as a permanent fix?

rainersigwald:
`dotnet watch` is a poster-child use-case for node reuse, so hopefully this is just temporary. It's a pretty reasonable guess if new MSBuild introduced flakiness, though.

Eilon:
Thanks @rainersigwald that's what I was figuring.

natemcmaster:
macOS tests still failing...investigating more.

natemcmaster:
Turns out the _flakiness_ is not related to node-reuse. I've split these changes into https://github.com/aspnet/DotNetTools/pull/419. 

There are issues with node-reuse, however, and they appear to be the result of https://github.com/Microsoft/msbuild/issues/3140. These failures don't currently show up in the PR checks because I'm still waiting for https://github.com/aspnet/BuildTools/pull/627. The failures appear to be indefinite hangs when invoking MSBuild from within another MSBuild process.

natemcmaster:
I'm going to close this for now without merging. It doesn't appear that disabling node reuse in dotnet-watch fixes the issue. It appears we'll need to apply some other kind of workaround based on the output of https://github.com/Microsoft/msbuild/issues/3140

PR #422, opened 2018-04-05
Ensure restore settings and package versions flow into test projects during CI builds

natemcmaster:
When MicrosoftNETCoreApp21PackageVersion is overridden during a prodcon build, all dotnet-watch functional tests fail because they cannot find the right feeds to restore the package from, so they rollforward to the latest Microsoft.NETCore.App on myget. This fails because a matching runtime for that new package is not available.

Changes:
 - capture all restore feeds used during build and generate restore settings correctly (current NuGet.config generation puts the file in the wrong place)
 - capture the Microsoft.NETCore.App package version used during build and generate test settings to ensure test projects restore the same version

@Eilon @muratg @DamianEdwards - ASK
This is a test-only change, no product changes, and should fix all failing tests in release/2.1 in ProdCon builds, such as http://aspnetci/viewLog.html?buildId=440648&tab=buildResultsDiv&buildTypeId=Releases_21Public_Win2012

FYI @ryanbrandenburg

### DIFF
--- a/test/dotnet-watch.FunctionalTests/Scenario/ProjectToolScenario.cs
+++ b/test/dotnet-watch.FunctionalTests/Scenario/ProjectToolScenario.cs
@@ -5,9 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Internal;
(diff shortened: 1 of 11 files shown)


test/dotnet-watch.FunctionalTests/Scenario/ProjectToolScenario.cs
```diff
@@ -28,21 +29,18 @@ public ProjectToolScenario()
 
         public ProjectToolScenario(ITestOutputHelper logger)
         {
+            WorkFolder = Path.Combine(AppContext.BaseDirectory, "tmp", Path.GetRandomFileName());
```
Eilon: I love `Path.GetRandomFileName`! This is the GOOD API!

PR #434, opened 2018-05-04
[dotnet-watch] Don't show false error messages

natemcmaster:
Don't show error when inner process exited because we killed it due to a file changing, or when the users requests cancellation via ctrl+c

*Before*

*With this change*

cc @glennc


natemcmaster:
@pakrym can you take a quick look? this is a minor change for 2.2

Issue #409, opened 2018-03-19
Flaky test: MultipleTriggers


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=422297&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR) with this error:

```
The collection contained 2 matching element(s) instead of 1.
   at Microsoft.DotNet.Watcher.Tools.FunctionalTests.FileWatcherTests.<>c__DisplayClass8_0.<MultipleTriggers>b__0(String dir) in /_/test/dotnet-watch.FunctionalTests/FileWatcherTests.cs:line 333
   at Microsoft.DotNet.Watcher.Tools.FunctionalTests.FileWatcherTests.UsingTempDirectory(Action`1 action) in /_/test/dotnet-watch.FunctionalTests/FileWatcherTests.cs:line 428
```

ryanbrandenburg:
Saw this again, [here's](http://aspnetci/httpAuth/app/rest/builds?locator=tag:FileWatcherWrong,sinceDate:201710010T000000-0001&fields=build(startDate,buildTypeId,status,branchName,webUrl,tags)) a list of these failures we've seen.

natemcmaster:
Added more output to the test log. Let me know next time it fails. The log should have more info then.

natemcmaster:
Thanks for investigating more. Looks like our test code is bad. Let's update the assertions to be less strict about how many events are raised.

natemcmaster:
Should be fixed in https://github.com/aspnet/DotNetTools/pull/435. I'll keep an eye on CI for future failures.

PR #435, opened 2018-05-08
Refactor MultipleTriggers test to better handle multiple events raised on the same file

natemcmaster:
Resolves https://github.com/aspnet/DotNetTools/issues/409

The test was failing on Assert.Single, but OnFileChanged may be raised multiple times, depending on what the System.IO APIs do under the hood.


natemcmaster:
FYI - running this test in a loop on my mac failed consistently within 2-3 loops. With this refactoring, it ran without failing for 10 minutes. Same on Windows.

PR #433, opened 2018-05-03
Upgrade to netcoreapp22

ryanbrandenburg:

PR #437, opened 2018-05-09
Add suppressions for NPV errors

pranavkm:


NuGetPackageVerifier.json
```diff
@@ -15,39 +15,76 @@
         ],
         "Exclusions": {
           "NEUTRAL_RESOURCES_LANGUAGE": {
+            "tools/netcoreapp2.1/any/System.Diagnostics.DiagnosticSource.dll": "Assembly is built by another project but bundled in our nupkg.",
```
pranavkm: More things show up when we build against the 4.6.0 packages.

pranavkm:
> are these new files present because of expected changes to what's in Microsoft.NETCore.App? 

That bit I'm not sure. I can see these files show by updating individual packages in the repo e.g. updating to System.Security.Cryptography.Cng causes `tools/netcoreapp2.1/any/System.Security.Cryptography.Cng.dll` to start appearing in the dotnet-dev-certs package. This does warrant further investigation, at the same time, I'd like to keep Update-Universe, currently broken by this check, going.

natemcmaster:
Can you try updating the projects in src/* to netcoreapp2.2? It looks like https://github.com/aspnet/DotNetTools/pull/433 only upgraded the tests and samples

pranavkm:
Ahh, I assumed they were meant to be left alone. I'll give that a try.

pranavkm:
Much better. 🆙 📅

pranavkm:
Getting this in to unblock the CI. We can revisit this if it ends up being problematic.

natemcmaster:
retro-:shipit:. Thanks

Issue #410, opened 2018-03-19
CTRL-C with dotnet watch in Linux or Mac takes 20 to 30 seconds


MichaelSimons:
_Copied from https://github.com/dotnet/dotnet-docker/issues/431 by @richlander_ 

I am running a Docker container per the instructions in #420. CTRL-C works for ASP.NET locally and for console apps in a container but not for ASP.NET apps in a container. I get this same behavior on Windows and macOS.

Command line:

```console
docker run --rm -it -p 8000:80 -v c:\git\dotnet-docker\samples\aspnetapp:/app/ -w /app/aspnetapp microsoft/dotnet-nightly:2.1-sdk dotnet watch run
```

Result:

```console
Hosting environment: Production
Content root path: /app/aspnetapp
Now listening on: http://[::]:80
Application started. Press Ctrl+C to shut down.
^CApplication is shutting down...
watch : Shutdown requested. Press Ctrl+C again to force exit.
^C
^C^C^C^C^C^C
^C^C^C^C^C^C^C^C
^C^C^C^C^C
```

/cc @natemcmaster @MichaelSimons @glennc

_Copied from https://github.com/dotnet/dotnet-docker/issues/431#issuecomment-374237841 by @MichaelSimons_

This appears to be tied to using `dotnet watch` in conjunction with ASP.NET.  CTRL-C works as expected when running an ASP.NET app directly within a container.

_Copied from https://github.com/dotnet/dotnet-docker/issues/431#issuecomment-374247175 by @MichaelSimons_

Upon further inspection, CTRL-C is working. It is just extremely slow.

_Copied from https://github.com/dotnet/dotnet-docker/issues/431#issuecomment-374384509 by @richlander_

I just repro'd that. It takes about 20s for me, on both macOS and Windows. Surprisingly, Windows containers don't seem to have this problem.

_Copied from https://github.com/dotnet/dotnet-docker/issues/431#issuecomment-374390354 by @natemcmaster_

I can repro, too. You can move this to aspnet/DotNetTools. I have a hunch it may be something in the "kill process tree" code, but I'm not sure.

muratg:
Ugh. cc @glennc

muratg:
Moving this out due to scheduling constraints.

natemcmaster:
[...]

PR #438, opened 2018-05-14
Dispatch cancellation on CTRL+C to avoid deadlocks in killing the process tree

natemcmaster:
Resolves https://github.com/aspnet/DotNetTools/issues/410.

This changes from calling CancellationTokenSource.Cancel inline with CTRL+C to dispatching to workaround https://github.com/dotnet/corefx/issues/29699 on macOS/Linux.

### DIFF
--- a/shared/CliContext.cs
+++ b/shared/CliContext.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -8,13 +8,12 @@ namespace Microsoft.Extensions.Tools.Internal
     public static class CliContext
     {
         /// <summary>
-        /// dotnet --verbose subcommand
+        /// dotnet -d|--diagnostics subcommand
         /// </summary>
         /// <returns></returns>
         public static bool IsGlobalVerbose()
         {
-            bool globalVerbose;
-            bool.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_CONTEXT_VERBOSE"), out globalVerbose);
+            bool.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_CONTEXT_VERBOSE"), out bool globalVerbose);
             return globalVerbose;
         }
     }
--- a/src/dotnet-watch/Internal/ProcessRunner.cs
+++ b/src/dotnet-watch/Internal/ProcessRunner.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -33,7 +33,7 @@ namespace Microsoft.DotNet.Watcher.Internal
             var stopwatch = new Stopwatch();
 
             using (var process = CreateProcess(processSpec))
-            using (var processState = new ProcessState(process))
+            using (var processState = new ProcessState(process, _reporter))
             {
                 cancellationToken.Register(() => processState.TryKill());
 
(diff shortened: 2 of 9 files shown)


src/dotnet-watch/Program.cs
```diff
@@ -121,7 +121,15 @@ private void OnCancelKeyPress(object sender, ConsoleCancelEventArgs args)
                 _reporter.Output("Shutdown requested. Press Ctrl+C again to force exit.");
             }
 
-            _cts.Cancel();
+            if (!Thread.CurrentThread.IsThreadPoolThread)
```
Tratcher: Is it ever a thread pool thread?

natemcmaster: It is during unit tests.

src/dotnet-watch/Internal/ProcessRunner.cs
```diff
@@ -95,29 +95,38 @@ private Process CreateProcess(ProcessSpec processSpec)
             }
         }
 
-        private class ProcessState : IDisposable
+        private sealed class ProcessState : IDisposable
```
halter73: Any reason for sealing private classes now?

shared/CliContext.cs
```diff
@@ -8,14 +8,13 @@ namespace Microsoft.Extensions.Tools.Internal
     public static class CliContext
     {
         /// <summary>
-        /// dotnet --verbose subcommand
+        /// dotnet -d|--diagnostics subcommand
```
halter73: Was this just wrong before?

src/dotnet-watch/Program.cs
```diff
@@ -121,7 +125,19 @@ private void OnCancelKeyPress(object sender, ConsoleCancelEventArgs args)
...
+            _userInterrupt.Cancel();
+
+            if (!Thread.CurrentThread.IsThreadPoolThread)
+            {
+                // Invoke the cancellation on the default thread pool to workaround https://github.com/dotnet/corefx/issues/29699
+                ThreadPool.QueueUserWorkItem(_ => _cts.Cancel());
+            }
+            else
+            {
+                _cts.Cancel();
```
halter73: Is there a scenario where QUWI doesn't work? I'd always cancel the CTS using QUWI for consistency.

src/dotnet-watch/Program.cs
```diff
@@ -121,7 +125,19 @@ private void OnCancelKeyPress(object sender, ConsoleCancelEventArgs args)
                 _reporter.Output("Shutdown requested. Press Ctrl+C again to force exit.");
             }
 
-            _cts.Cancel();
+            // always fire user interupt inline because this used to determine if the process
+            // is exiting due to user interaction or not
+            _userInterrupt.Cancel();
```
halter73: Can you explain why there are two tokens now given that you're cancelling them one after the other? From what I can tell this is the only place either token is canceled.

shared/CliContext.cs
natemcmaster: This was `dotnet --verbose` was renamed after the project.json -> msbuild conversion. Just updating the doc comment to be accurate.

src/dotnet-watch/Program.cs
natemcmaster: I was getting a race condition in the unit tests that fakes the CTRL+C behavior.

src/dotnet-watch/Program.cs
```diff
@@ -16,17 +16,21 @@ public class Program : IDisposable
     {
         private readonly IConsole _console;
         private readonly string _workingDir;
-        private readonly CancellationTokenSource _cts;
         private IReporter _reporter;
 
+        // Use this token to check if the user has pressed CTRL+C
+        private readonly CancellationTokenSource _userInterrupt = new CancellationTokenSource();
```
halter73: Does this ever get disposed?

natemcmaster: No. Should it?

src/dotnet-watch/Program.cs
natemcmaster: Reason for adding a separate token is to not regress on https://github.com/aspnet/DotNetTools/pull/434. Basically, when the watchee (not the watcher) exits, we want to know if it did so because the user invoked ctrl+c or a file changes. By changing to dispatching _cts.Cancel(), it appears the order in which some tasks complete changes slightly and regressed https://github.com/aspnet/DotNetTools/pull/434

https://github.com/aspnet/DotNetTools/pull/438/files#diff-7895c5cf5e852895b5c8be62b850d40fR75

src/dotnet-watch/Program.cs
halter73: I generally try to and you do for _cts, so it's probably better to be consistent. It matters mostly if you creating CTS doesn't get canceled prior to being dereferenced in order to clean up the registrations.

src/dotnet-watch/Program.cs
halter73: No big deal either way. I'd probably create a helper method that verifies either that the token is already canceled or becomes canceled in 30 seconds.

src/dotnet-watch/Program.cs
halter73: That's weird. If you're trying to determine whether or not cancelling the CTS triggered the process exit (or is that not what you're trying to determine?), wouldn't the token have to be canceled by the time the process exits in order for that to ever be true even if the cancellation was dispatched?

src/dotnet-watch/Internal/ProcessRunner.cs
natemcmaster: IIUC it's a very minor jit/clr perf improvement, but I'm not completely sure. I've been asked by others to add 'sealed' to private classes in other places, so I assume there are reasons for it.

src/dotnet-watch/Program.cs
natemcmaster: actually, tested this more. The race condition was a different problem. Updated to always dispatch now.

src/dotnet-watch/Program.cs
natemcmaster: good point. I played with this some more and think the original cts is good enough. Removed the duplicate cts.

PR #443, opened 2018-05-23
Add support for DOTNET_WATCH_ITERATION

### LINKED ISSUE #387: Add counter to dotnet watch
Make it possible to detect from the inner process how many times the process has been restarted.

Ideas:
set the environment variable DOTNET_WATCH_ITERATION to an integer, starting at 0

Example:
```
> dotnet watch run
Starting: dotnet run, DOTNET_WATCH_ITERATION=0
(Change to file, dotnet-run is killed)
Retarting: dotnet run, DOTNET_WATCH_ITERATION=1
```

cc @danroth27

natemcmaster:
This environment variable can be used to determine how many types the inner command has been re-launched

Resolves #387

### DIFF
--- a/src/dotnet-watch/CommandLineOptions.cs
+++ b/src/dotnet-watch/CommandLineOptions.cs
@@ -53,6 +53,10 @@ Environment variables:
   DOTNET_WATCH
   dotnet-watch sets this variable to '1' on all child processes launched.
 
+  DOTNET_WATCH_ITERATION
+  dotnet-watch sets this variable to '0' and increments by one each time
+  the inner command is restarted.
+
 Remarks:
   The special option '--' is used to delimit the end of the options and
   the beginning of arguments that will be passed to the child dotnet process.
--- a/build/repo.props
+++ b/build/repo.props
@@ -2,9 +2,6 @@
   <Import Project="dependencies.props" />
 
   <ItemGroup>
-    <DotNetCoreRuntime Include="2.0.0" />
-    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp20PackageVersion)" />
-    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp21PackageVersion)" />
     <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp22PackageVersion)" />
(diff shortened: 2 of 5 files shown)


src/dotnet-watch/CommandLineOptions.cs
```diff
@@ -53,6 +53,10 @@ public static CommandLineOptions Parse(string[] args, IConsole console)
   DOTNET_WATCH
   dotnet-watch sets this variable to '1' on all child processes launched.
 
+  DOTNET_WATCH_ITERATION
+  dotnet-watch sets this variable to '0' and increments by one each time
```
muratg: Perhaps 1 is a better starting count?

src/dotnet-watch/CommandLineOptions.cs
```diff
@@ -53,6 +53,10 @@ public static CommandLineOptions Parse(string[] args, IConsole console)
...
+  the inner command is restarted.
```
muratg: "Inner command"?  Docs leaking implementation details?

src/dotnet-watch/CommandLineOptions.cs
natemcmaster: I'm fine either way. starting at 1 was my first choice too, but then I probably overthought this on account of all those "arrays start at 1" jokes on Reddit. I've updated the PR to start the counter at 1.

src/dotnet-watch/CommandLineOptions.cs
natemcmaster: What about this?

> dotnet-watch sets this variable to '1' and increments by one each time a file is changed and the command is restarted.

natemcmaster:
I chatted with @glennc in the hallway. He said he's okay with it, though we should probably consider this altogether with a design for launching the browser from `dotnet run`. Browser-launching was the original motivation for me adding this. We want to be able to launch the browser from `dotnet run`, but we also don't want to re-launch the browser everytime you change a .cs file. Adding a counter gives the browser-launching thingy the ability to make this call without dotnet-watch needing to know about how the browser-launching thingy will work.

Issue #166, opened 2016-09-20
Add user-secret init command


natemcmaster:
_From @Tratcher on December 10, 2015 18:37_

When you try to use user-secret on a new project it fails with:
`critical: Command failed : Missing 'userSecretsId' in 'C:\git\Universe\Security\samples\JwtBearerSample\project.json'`

There should be an `init` command to generate the userSecretsId in the project.json.

_Copied from original issue: aspnet/UserSecrets#53_

natemcmaster:
Relevant work on CLI: https://github.com/dotnet/cli/issues/4746

natemcmaster:
Remarking for triage: as this should be trivial now that preview5 of CLI has built infrastructure for this. `dotnet user-secrets init` could easily shell out to `dotnet add package Microsoft.Extensions.Configuration.UserSecrets`

muratg:
Depends on https://github.com/dotnet/cli/issues/5300

liamdawson:
@muratg would you be interested in a more naive implementation until dotnet/cli#5300 is complete?

natemcmaster:
@liamdawson depends on how it was implemented. Can you share a brief description of how you would implement this?

liamdawson:
Certainly.

* Add a command called `init`
    * Support a `--force` flag to force resetting the secrets ID
* `init` uses the `MsBuildProjectFinder` class to find the project file
* Parse the project file as XML, check if any of the `<PropertyGroup>` elements contain a `<UserSecretsId />`
    * If not, add one with a new `Guid` to the first `<PropertyGroup>`
    * If exists and not `--force`, exit with an error
    * If exists and `--force`, set a new `Guid` on the existing `<UserSecretsId />`

I'm sure I'm missing intricacies of the MS Build system, but I think that would be sufficient for the typical `dotnet new` project, at least?

natemcmaster:
Yes, I think its safe to treat the file as XML. _Technically,_ it would be better to use the API in [Microsoft.Build.Construction](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.construction?view=netframework-4.7) namespace, but we have avoided adding MSBuild references to our CLI tools for a handful of reasons. XML should be OK for this purpose.

A few additions:
[...]

PR #500, opened 2018-10-13
Add an init command for adding a UserSecretsId

liamdawson:
Adds an init command for adding a UserSecretsId

Addresses #166

### DIFF
--- a/src/dotnet-user-secrets/Internal/InitCommand.cs
+++ b/src/dotnet-user-secrets/Internal/InitCommand.cs
@@ -1,13 +1,16 @@
 using Microsoft.Extensions.CommandLineUtils;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Internal
 {
     internal class InitCommand : ICommand
     {
-        public string _overrideId { get; }
+        public string OverrideId { get; }
+        public string ProjectPath { get; }
+        public string WorkingDirectory { get; private set; } = Directory.GetCurrentDirectory();
 
         public static void Configure(CommandLineApplication command, CommandLineOptions options)
         {
@@ -16,18 +19,31 @@ namespace Microsoft.Extensions.SecretManager.Tools.Internal
 
             command.OnExecute(() =>
             {
-                options.Command = new InitCommand(options.Id);
+                options.Command = new InitCommand(options.Id, options.Project);
             });
         }
 
-        public InitCommand(string id)
+        public InitCommand(string id, string project)
         {
-            _overrideId = id;
+            OverrideId = id;
+            ProjectPath = project;
+        }
+
+        private static string ResolveProjectPath(string name, string path)
+        {
+            var finder = new MsBuildProjectFinder(path);
+            return finder.FindMsBuildProject(name);
         }
 
         public void Execute(CommandContext context)
         {
-            throw new NotImplementedException();
+            var projectPath = ResolveProjectPath(ProjectPath, WorkingDirectory);
+        }
+
+        public void Execute(CommandContext context, string workingDirectory)
+        {
+            WorkingDirectory = workingDirectory;
+            Execute(context);
         }
     }
 }
--- a/src/dotnet-user-secrets/Internal/InitCommand.cs
+++ b/src/dotnet-user-secrets/Internal/InitCommand.cs
@@ -10,11 +10,10 @@ using Microsoft.Extensions.CommandLineUtils;
 
 namespace Microsoft.Extensions.SecretManager.Tools.Internal
 {
-    public class InitCommand : ICommand
+    // Workaround used to handle the fact that the options have not been parsed at configuration time
+    public class InitCommandFactory : ICommand
     {
-        public string OverrideId { get; }
-        public string ProjectPath { get; }
-        public string WorkingDirectory { get; private set; } = Directory.GetCurrentDirectory();
+        public CommandLineOptions Options { get; }
 
         internal static void Configure(CommandLineApplication command, CommandLineOptions options)
         {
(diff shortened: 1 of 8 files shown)


liamdawson:
There are a couple of things I would like some help fixing appropriately:

1. Any test cases I'm missing? e.g. I feel I am missing test cases for invalid/missing project files for `init`, as they won't return the same output as they do for other commands right now.
2. As the above point, some error cases (invalid/missing project file) won't return output consistent with other subcommands - what's the best DRY way to solve that?
3. How should I indicate a command failure for https://github.com/aspnet/DotNetTools/pull/500/files#diff-0207ee8b1aa083ec78304322c8b620c7R65 ?
4. How do I add a localized error for "Project already has a UserSecretsId"? Adding it to the `resources.resx` didn't add the functions in the `designer.cs` that exist for the other resource entries.

natemcmaster:
Hi @liamdawson, thanks for doing this work! I'm going to help you merge this PR in. Unfortunately, we've missed the deadline for features in 2.2, so this PR will need to target 3.0 instead. I'm going to re-target your PR to the master branch and then review the code changes.

Thanks,
Nate

src/dotnet-user-secrets/Internal/InitCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+            // Load the project file as XML
+            var projectDocument = XDocument.Load(projectPath);
+
+            // Confirm a UserSecretsId isn't already set
+            if (projectDocument.XPathSelectElements("//UserSecretsId").Any())
+            {
+                // TODO: i18n
+                context.Reporter.Error($"MSBuild project '{projectPath}' already contains a UserSecretsId.");
+                // TODO: correct error reporting approach
+                throw new NotImplementedException();
```
natemcmaster: IMO tools should not fail if there action would be a no-op. They can just exit early. I think this code should be tweaked a little so we get the following behaviors.

Was `--id` given?  | Project contains `<UserSecretsId>` already? | Action
---|---|--
No | Yes | Message: "Project is already initialized", exit code 0
No | Yes | Generate GUID, add `<UserSecretsId>` element, exit code 0
Yes | No | Add `<UserSecretsId>` element, exit code 0
Yes | Yes | Overwrite `<UserSecretsId>` element, exit code 0

Reasons to fail:
* the ID specified with `--id` is invalid.
* Project file could not be found
* Project file is malformed (not proper XML)

src/dotnet-user-secrets/Internal/InitCommand.cs
```diff
@@ -0,0 +1,99 @@
+using Microsoft.Extensions.CommandLineUtils;
```
natemcmaster: Two code formatting nit-picks and all new *.cs files:
* Can you add the copyright header?
* Can you sort using with System.* namespaces first? I wish VS enforced this better... :-/ https://github.com/aspnet/DotNetTools/blob/581b3160f422180b3411d84b4458c10128c8cacb/.editorconfig#L12-L14

src/dotnet-user-secrets/Internal/InitCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+            var propertyGroup = projectDocument.Root.DescendantNodes()
+                .FirstOrDefault(node => node is XElement el
+                    && el.Name == "PropertyGroup"
+                    && el.Attributes().All(attr =>
+                        attr.Name != "Condition")) as XElement;
+
+            // No valid property group, create a new one
+            if (propertyGroup == null)
+            {
+                propertyGroup = XElement.Parse(@"<PropertyGroup></PropertyGroup>");
```
natemcmaster: Does this work? Parsing seems like a little much.
```
propertyGroup = new XElement("PropertyGroup");
```

src/dotnet-user-secrets/Internal/InitCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                projectDocument.Root.AddFirst(propertyGroup);
+            }
+
+            // Add UserSecretsId element
+            var userSecretsElement = XElement.Parse(@"<UserSecretsId></UserSecretsId>");
```
natemcmaster: Does this work?
```
var userSecretsElement = new XElement("UserSecretsId", newSecretsId);
```

src/dotnet-user-secrets/Program.cs
```diff
@@ -71,6 +71,12 @@ internal int RunInternal(params string[] args)
 
             var reporter = CreateReporter(options.IsVerbose);
 
+            if (options.Command is InitCommand initCmd)
+            {
+                initCmd.Execute(new CommandContext(null, reporter, _console), _workingDirectory);
+                return 0;
```
natemcmaster: In retrospec, Command.Execute should probably have returned `int` so the command can set an exit code. I'd be happy to also take a PR that made this change.

liamdawson:
Thanks for the thorough feedback @natemcmaster, I'll see to that soon.

Could I also get a pointer on how to add messages via `resources.resx`? I'll need to add a couple, and adding them in VS didn't seem to update the appropriate designer file (which says it is auto-generated).

Thanks again!

natemcmaster:
Re: resx files.

Edit the .resx file in VS (or an xml editor) and then run `dotnet msbuild -restore -t:Resx` from command line.

natemcmaster:
Sorry, missed a piece: run `dotnet msbuild -restore -t:Resx` _in the `src/dotnet-user-secrets/` project dir._

src/dotnet-user-secrets/Program.cs
liamdawson: Okay - I looked at that but ran into the fact the interface had the expectation. At first glance though, I don't know how I'd safely modify the interface to allow a status code return without causing a breaking change (but that's a separate conversation).

src/dotnet-user-secrets/Internal/InitCommand.cs
liamdawson: Right, I didn't actually try that because I was a bit superstitious about `XName`. Will take a look later.

src/dotnet-user-secrets/Program.cs
natemcmaster: You have my blessing to make a breaking change on `ICommand`. For projects like this, the `ICommand` type is an internal implementation detail of these console tools. No one should be compiling an app against `dotnet-user-secrets.dll`.

liamdawson:
I'll open another PR to cover the status code returns on `ICommand`, otherwise I believe I've addressed all your PR feedback (and have removed changes to korebuild and csproj files)

src/dotnet-user-secrets/Program.cs
liamdawson: The `ICommand` in this instance is from CommandLineUtils, so not part of this tool - do you still want that change?

natemcmaster:
@liamdawson ACK saw your comments. I'll take a closer look at this in a few days. Currently working on time-sensitive issues. Please ping me on this PR  if I don't surface for air by Thurs. Nov. 1.

natemcmaster:
Thanks for making the requested changes. I'll bring this in once CI is green (I've just re-triggered a build).

PR #512, opened 2018-11-05
Reduce flakiness in tests caused by the OS re-using the PID on test processes

natemcmaster:
Resolves https://github.com/aspnet/DotNetTools/issues/511.

We were using PID to assert that dotnet-watch had restarted processes. The OS does not guaranteed the uniqueness of PID, so I've changed tests to use PID + process start time to distinguish between processes.


natemcmaster:
:bell:

muratg (approved):

